PR #7, opened 2015-01-15
Chromosome split

dozy:
changes for y chromosome split
prevented p4 processing of plexes which are tagged as phix


dozy:
This shouldn't be released until a new p4 release is done (which will include the necessary templates to handle ysplit).

PR #25, opened 2015-02-10
Mlwh driver4samplesheet

mgcam:


dkj:
Agreed with Marina we'll merge directely into devel (with changes to work with existing LIMS adapters as necessary)

PR #29, opened 2015-02-20
PB_cal_bam pipeline renamed to central

mgcam:


mgcam:
Test t/10-cache.t fails (in npg::samplesheet) if no connection to tracking database available. This is not a new issue. Since we'd like to rewrite extended samplesheet generation so that no access to tracking database is made, I decided not to fix the test.

PR #41, opened 2015-03-11
pull hot fix (release 47.9.1) from master to devel

mgcam:


mgcam:
cannot be merged automatically

PR #65, opened 2015-04-17
Rel 48.4

dozy:


srl147:
this doesn't fix the problem as all tags have no reference

PR #66, opened 2015-04-17
Add 'lowload' lfs queue

jenniferliddle:


dkj:
warehouse updating jobs as well?

jenniferliddle:
As far as I can see, warehouse updating does not use LSF, the jobs run directly on the head nodes via the crontabs.

jenniferliddle:
Got it. New lowload queue now used for warehouse loader and fastqcheck loader

PR #84, opened 2015-06-01
gclp: force p4 style alignment

mgcam:
and fix the daemon test to run correctly against pending changes in https://github.com/wtsi-npg/ml_warehouse/pull/40


dozy:
'gclp flag is set to true' test in t/50-harold_analysis_runner.t fails as noted in the pull request comment

PR #100, opened 2015-06-30
allow p4 to be used where no alignment is specified for target but hu…

dozy:
…man split is (contains_nonconsented_human)

Note: this still needs tests to be added


dkj:
I'm happy with this - there's a conflict resolved version, with extra QC changes required for #106, in #107

PR #105, opened 2015-07-09
wh loader - be verbose, load mlwh at the end of analysis

mgcam:


lib/npg_pipeline/archive/file/generation.pm
```diff
@@ -1,15 +1,7 @@
...
-use Moose::Meta::Class;
-use Carp;
-use English qw{-no_match_vars};
-use File::Spec;
-
-use npg_common::roles::run::lane::file_names;
-use npg_tracking::glossary::tag;
-use npg_tracking::glossary::lane;
-use npg_tracking::glossary::run;
+use File::Spec qw/catfile/;
```
dkj: ```use File::Spec::Functions qw(catfile);``` to work without ```File::Spec->``` ?

dkj:
Fine with this - but needs ```use File::Spec::Functions``` change - this is in combo PR #107

PR #106, opened 2015-07-14
use 'subset' option of the bam_flagstast autoqc result

mgcam:


dkj:
Fine with this - have a conflict resolved post #105 merge in #107

PR #107, opened 2015-07-17
Combination merge for warehouse, QC, human split with no target alignment

dkj:


dozy:
Note: the human split with no target alignment code depends on a new release of p4 (current release is v0.16.4)

PR #122, opened 2015-11-12
Bam and cram creation threading

dkj:
improve the efficiency of the second half of P4 based seq_alignment jobs


dkj:
(Testing on sf8_1, with data 18072_1) Wasted CPU fraction reduced from 39% to 33% (down to 17% with wtsi-npg/p4#126).

dkj:
Made merge ready, but also includes #127

PR #124, opened 2015-11-18
seq_alignment - P4 and new bwa for older chemistrie…

jenniferliddle:
…s & forcing mem for alt references


dkj:
Is there a test for the alt_genome method? If not can one be added please? I don't understand how it's working... I thought you need get the bwa0_6 path/prefix then append (something like) ".alt" and test for file existence.

dkj:
We need to force bwa mem for gclp - but this can come in via a separate PR....

PR #125, opened 2015-12-04
Study specific env prework

mgcam:
Previminary work for study-specific analysis environment. A common parent class for pipeline daemons is introduced. Previously the archival daemon inherited from the analysis one. Since the analysis daemon will be extended, it's better to have common functionality in a parent module. 

Extract from Changes file
----------------------------------
- require minimum version 5.10 for perl
 - add study analysis configuration accessor
 - simpler name for the archival daemon module
 - parent class for pipeline daemons
 - consistent behaviour of the archival and analysis daemons
   when LIMs data are not available in the ml warehouse -
   the run is skipped
 - Log::Log4perl logger is used in pipeline daemons

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -20,6 +20,7 @@ my $builder = $class->new(
          'dist'                => { COMPRESS => 'gzip', SUFFIX => 'gz', },
 
          'requires'            => {
+          'perl'                         => 5.010,
           'autodie'                      => 0,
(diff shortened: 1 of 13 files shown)


mgcam:
More features will be added

mgcam:
This code is included in https://github.com/wtsi-npg/npg_seq_pipeline/pull/131

PR #127, opened 2015-12-09
GCLP attribute

dkj:
add attribute "gclp" common to analysis and archival scripts
ensure we use bwa mem for GCLP


dkj:
Ensures we use bwa mem for GCLP consistent with previously following #124

lib/npg_pipeline/base.pm
```diff
@@ -240,8 +257,13 @@ Boolean decision to force on P4 pipeline usage
 has q{force_p4}  => (
   isa           => q{Bool},
   is            => q{ro},
-  documentation => q{Boolean decision to force on P4 pipeline usage},
+  lazy_build    => 1,
+  documentation => q{Boolean decision to force on P4 pipeline usage, default true iff GCLP},
```
mgcam: typo: iff -> if

dkj: Ummm: I meant [if and only if](https://en.wikipedia.org/wiki/If_and_only_if). I agree "if" would suffice for this description though.

dkj:
@jenniferliddle always including 'pipeline_name@ i.e. class suffix, in generated function name now ensures we default to correct config file for npg_pipeline_central --gclp. It also tightens the config file selection when using functions like --olb and --qc_run.

PR #131, opened 2015-12-15
pipeline switches in daemons, ml warehause as source of LIMs data

mgcam:


mgcam:
As an added bonus, runs with batch unreleased (no data in ml iseq_warehouse flowcell table) will not go into analysis.

mgcam:
Comments after a preliminary review by djk:
1. Using gclp glag as a pipeline option is not compatible with teh current analysis pipeline on the gclp cluster; we shoudl maintain compatibility.
2. create dry-run option for the pipeline daemons so that they can be tested against libe dbs.

lib/npg_pipeline/daemons/base.pm
```diff
@@ -33,37 +29,61 @@ Readonly::Scalar my $NO_LIMS_LINK => -1;
...
+  isa        => q{Log::Log4perl::Logger},
+  is         => q{ro},
+  metaclass  => 'NoGetopt',
+  default    => sub { Log::Log4perl->get_logger() },
+);
+
+around 'log' => sub {
+   my $orig = shift;
+   my $self = shift;
+   $self->logger->info('"log" method is depricated');
```
dkj: deprecated

mgcam: > + $self->logger->info('"log" method is depricated');
>
> deprecated
>
Thank you, fixed.




-- 

PR #133, opened 2015-12-17
use bwa_aln_se for non paired reads

jenniferliddle:


dkj:
Test case for single read, rather than paired, would be useful.

PR #139, opened 2016-01-15
purge carriage returns (as well line feeds) from study descriptions

dkj:
for RG header records (xml lims driver has previously done this as part of XML parsing)


dkj:
Likely duplicate of work by @mgcam and @srl147

PR #142, opened 2016-01-20
add qc check rnaseqc

rube14:
Branch includes the addition of new Broad Institute's RNA-SeQC QC check for RNA libraries (human and mouse only)


mgcam:
First we have to look at the implementation of this new check and incorporate it into the npg_qc package.  Then we can have a look at this pull request.

mgcam:
This pull request has not been reviewd yet.

data/config_files/function_list_central_qc_run.yml
```diff
@@ -27,6 +27,7 @@
 - qc_genotype
 - qc_verify_bam_id
 - qc_upstream_tags
+- qc_rna_seqc
```
dkj: Don't think we should add it in the "qc" runs. These are the paired 25bp runs used for quantification of libraries (I'll submit a separate PR to remove other inappropriate QC checks).

PR #144, opened 2016-01-25
bug fix: runfolder_path needs to be available for archival daemon

mgcam:
based on work by @dkj , see https://github.com/wtsi-npg/npg_seq_pipeline/pull/143


mgcam:
Tested both teh srchival and analysis daemon scripts in the dry_run mode

PR #147, opened 2016-02-15
P4 stage1 analyses

dozy:


dkj:
Drop max number of slots used to 4? (Current usage seems lower...).

PR #149, opened 2016-02-22
use 'purpose' field to decide if qc_run

jenniferliddle:


lib/npg_pipeline/daemon.pm
```diff
@@ -200,6 +200,7 @@ sub check_lims_link {
   $lims->{'id'} = $batch_id;
   if ($fcell_row) {
     $lims->{'gclp'} = $fcell_row->from_gclp;
+    $lims->{'qc_run'} = ($fcell_row->purpose eq 'qc') ? 1 : undef;
```
mgcam: $fcell_row->purpose might be undefined

jenniferliddle: Fixed

mgcam:
run_spider() method in npg_pipeline::pluggable::harold uses the qc_run flag to switch between the new and old wh drivers for generating a samplesheet. This has to be fixed.

PR #165, opened 2016-06-16
only single parallelisation config file required

dkj:
@dozy What do you think?


mgcam:
On 16/06/2016 14:24, dkj wrote:
> What do you think?
Seems fine to me.


-- 

PR #166, opened 2016-06-21
P4 splice prune

dozy:
use the new node splicing/pruning feature in p4 to handle unplexed lanes (p4_stage1_analysis) and "no target alignment" (seq_alignment)


dkj:
```vtfp.pl``` is no longer launched with ```-s``` option - okay?

PR #167, opened 2016-06-22
Move runs to "analysis on hold" if 10xGenomics/Chromium library types

dkj:


dkj:
@srl147 points out that we now need to deal with things on a lane basis....

mgcam:
Discussed with @srl147. If a subset of lanes is analysed, the run will not go automatically to 'archival pending'. If moved manually, an attempt to archive missing lanes will be done, so, most likely, the archival pipeline will fail.

PR #176, opened 2016-08-25
fixes to satisfy autoqc objects' constraints in a newer version

mgcam:
This fixes the tests for creating autoqc jobs. However, in reality the can_run method will always give an error and, therefore return 1 because when the pipeline script is run the LSF variables are not defined and, therefore, the qc objects will get non-existing directories as both qc_in and qc_out (lane number will be expressed via an LSF job index). Therefore, either a directory check should be delayed till the check is run or we should pass an archive directory to the constructor when we need the output on can_run. The latter seem to be preferable to me. There also might be a problem in passing a reference attr to all checks if some of the checks have a strict constructor. Currently only qxYield. Probably the strict constructor has to be dropped. When the checks are run via a script, extra args are not a problem.

### DIFF
--- a/lib/npg_pipeline/archive/file/qc.pm
+++ b/lib/npg_pipeline/archive/file/qc.pm
@@ -165,9 +165,8 @@ sub _can_run {
   load_class($p);
 
   my $init_hash = {
-      path      => $qc_in,
+      qc_in     => $qc_in,
       position  => $position,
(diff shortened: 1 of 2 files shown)


mgcam:
Replaced with https://github.com/wtsi-npg/npg_seq_pipeline/pull/177

PR #184, opened 2016-09-30
Select an appropriate LIMS driver when loading data into iRODS.

keithj:
This is a first pass - I'm not sure that I have the function lists correct.

### DIFF
--- a/data/config_files/function_list_central.yml
+++ b/data/config_files/function_list_central.yml
@@ -28,5 +28,5 @@
 - qc_rna_seqc
 - run_analysis_complete
 - update_ml_warehouse
-- archive_to_irods
+- archive_to_irods_samplesheet
 - run_qc_review_pending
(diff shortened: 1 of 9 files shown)


dkj:
"job_name_prefix" is not used for new publish_illumina_run jobs - fallout from not using ```clone_with_attributes``` perhaps? Make me wonder are we missing anything else?

dkj:
The driver type is not propagating to the command in the LSF job. In the npg_pipeline_central log I see:
```
The new() method for npg_pipeline::archive::file::to_irods expects a hash reference or a key/value list. You passed an odd number of arguments at /nfs/users/nfs_d/dj3/repos/npg_seq_pipeline/install/51.2-46-gc6fff8c/bin/../lib/perl5/npg_pipeline/pluggable/harold/post_qc_review.pm line 85.
```

keithj:
The job_name_prefix and driver_type are being passed thanks to the reinstated magic.

PR #185, opened 2016-09-30
fix i2b and decode implementation selection

dozy:
fix facility to specify i2b and decode implementations (java/bambi) in p4 stage1 analysis via general_values.ini, and upgrade tests


dkj:
Only caveat is that to get 3 index read support @srl147 is likely to drop Illumina2bam use (for initial bcl to bam step anyway).

dozy:
The default i2b and decode implementation is already bambi in devel. This pull request will allow reversion to Illumina2bam if needed.

PR #188, opened 2016-10-14
Log4perl emulation of the current pipeline logging.

keithj:
This replaces the original log role with the one from DNAP utilities, which provides a Log4perl logger and some convenience methods.

As all the original log statements were at the same "level" I've made some judgement calls about what should be `INFO`, `WARN` and `ERROR`. These may need some adjustment. Where I can, I've replaced `croak` with `logcroak`.

I've done my best to reproduce the current logging behaviour, including its idiosyncracies e.g. logging to STDOUT.


mgcam:
Have not looked at any of the old/new code, writing to document an 
important feature of the pipeline which should be kept. While in some 
cases logging to stdout might have been accidental, in other cases it 
might have been deliberate. The pipeline script is called either by the 
pipeline daemon or manually. If an error occurs, it is caught, necessary 
clean-up is performed (the jobs that have been already created are 
killed), and then, most likely, the error is re-thrown. The information 
about this error should reach the caller, ie be logged in the daemon's 
log or, in case of a manual scenario, printed to console. Since stderr 
is diverted to the pipeline's own log, writing to stdout might have been 
the only way to make the error visible to the caller.

On 14/10/2016 12:03, Keith James wrote:
>
> I've done my best to reproduce the current logging behaviour, 
> including its idiosyncracies e.g. logging to STDOUT.
>



-- 

lib/npg_pipeline/analysis/split_bam_by_tag.pm
```diff
@@ -20,17 +20,17 @@ sub generate {
...
-    $self->log( qq{Run $id_run is not multiplex run and no need to split} );
+    $self->warn(qq{Run $id_run is not multiplex run and no need to split});
     return;
   }
 
   my %positions = map{$_=>1} $self->positions();
   my @indexed_lanes = grep {$positions{$_}} @{$self->multiplexed_lanes()};
   if ( ! @indexed_lanes ) {
-    $self->log( q{None of the lanes for analysis is multiplexed} );
+    $self->wanr(q{None of the lanes for analysis is multiplexed});
```
mgcam: warn?

mgcam:
Changes suggested in https://github.com/keithj/npg_seq_pipeline/pull/4 to emulate the current situation more closely.

One remaining difference: the original logger redirected stderr to the log file, so stderr output of all modules went to a log. The suggested version logs explicitly. Therefore, warnings from the pipeline and other modules and warnings/errors from Linux commands are not captured. They will either go a the pipeline daemin log or, in case of manual invocatio, to console. We might need to redirect stderr to the same log in scripts themselves.

PR #190, opened 2016-11-02
Add Travis scripts and minimal code changes to make travis work

jenniferliddle:


.travis.yml
```diff
@@ -0,0 +1,34 @@
+sudo: required
+
+language: perl
+
+perl:
+  - "5.16"
+
+addons:
+  postgresql: "9.3"
```
dkj: Is this required? Also, can we ditch the Jansson, Baton, iRODS stuff below?

t/10-cache.t
```diff
@@ -87,6 +87,8 @@ for my $type (qw/warehouse mlwarehouse/) {
   is( scalar @{$clims}, 2, 'two lims objects returned');
   is( $clims->[0]->driver_type, 'ml_warehouse', 'correct driver type');
 
+  SKIP: {
```
dkj: We should only skip if we cannot find the relevant old warehouse module - we've not purged it from our deployments yet. So some ```eval require``` in the test of whether to skip or not perhaps?

PR #206, opened 2016-12-20
Chain execution of RNA-SeQC qc check to the vtfp/viv alignment cmd for RNA-Seq libraries

rube14:
- Chained execution of RNA-SeQC to the vtfp/viv alignment cmd for RNA-Seq libraries only:
- Entries for qc check rna_seqc removed from central function and parallelisation lists.
- Code that created rna_seqc-specific directories has been removed as this is now handled by the check itself using qc_out arg.


mgcam:
I wonder whether it would be easier to have a default in the check itself for the output directory. The check could create a folder inside the directory where the json file goes. The name of the directory could be the root name of the result file. This way (a) the check would be forward compatible with a rapid run merge pipeline that is under development and (b) a change to the pipeline will be minimal. A change to the check will be very small as well since all needed values (names) are already available in the result object. 

Simultaneously we could have added an extra attribute for the result object (just for the file-based object) to store an abs path to this folder. Would work as long as the folder is not moved. Not brilliant, but the currently proposed solution - resolving a path relative to Latest Summary link - does take into account that re-analysis is often done without repointing the Latest Summary link. It would be much better to resolve the location of the folder relative to the location of the json file, but I do no have a good solution for this at the moment.

lib/npg_pipeline/archive/file/generation/seq_alignment.pm
```diff
@@ -193,6 +193,7 @@ sub _lsf_alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity
   my $spike_tag;
   my $input_path= $self->input_path;
   my $archive_path= $self->archive_path;
+  my $archive_qc_path = File::Spec->catdir($self->archive_path, q{qc});
```
mgcam: Is this variable being used?

rube14: It is not. Now fixed. Thanks.

mgcam:
This pull request looks fine to me, conflicts in Changes file should be resolved.

PR #213, opened 2017-04-12
supply --restart_file option to the iRODS publisher script

mgcam:
... and derive the name of this file from LSF job id so that the restart file is pinned to a specific LSF job.


srl147:
can your also add

  --max-errors ####

I suggest we start with a value of 10, via a new entry in the pipeline configuration file

PR #217, opened 2017-05-11
recalibration and olb removed

mgcam:


mgcam:
Coding is finished.
Planning to create a special dated directory for running the pipeline with this change.

PR #222, opened 2017-05-24
ysplit to use bambi

dozy:
change split by chromosome (Y split) parameters to bambi (remove illumina2bam)


dozy:
depends on https://github.com/wtsi-npg/p4/pull/182

PR #236, opened 2017-07-21
no alignments for chromium libraries

srl147:


lib/npg_pipeline/archive/file/generation/seq_alignment.pm
```diff
@@ -243,7 +243,7 @@ sub _lsf_alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity
                     $l->separate_y_chromosome_data    ? q(yhuman) :
                     q();
 
-  my $do_target_alignment = ($self->_ref($l,q(fasta)) and $l->alignments_in_bam);
+  my $do_target_alignment = ($self->_ref($l,q(fasta)) and $l->alignments_in_bam and not ($l->library_type && $l->library_type =~ /Chromium/smx));
```
dkj: Minor reference for consistent logic operators i.e. using `and` rather than `&&` to match existing usage.

PR #251, opened 2018-02-02
pipeline function dependencies as DAG

mgcam:
Also removes unused functions and modules.
Tested on LSF for a MiSeq run.


mgcam:
The code in this pr is half way towards what we eventually need. The final version is available in https://github.com/wtsi-npg/npg_seq_pipeline/pull/264. All code in this pr (251) is part of the final version (pr 264). We decided not to release this intermidiate version as the final one is practically ready; our time wil be better spend validating the final version.

PR #276, opened 2018-05-10
single log directory for all jobs with per-function subdirectories

mgcam:


mgcam:
based on https://github.com/wtsi-npg/npg_seq_pipeline/pull/275

PR #295, opened 2018-06-01
Minimap p4s2

dozy:


dozy:
- attempt to determine non-RNA and non-GbS target alignment method by parsing the lims reference genome name
- switch off target alignment for NovaSeq tag 0
- correct splice directives for spike tag processing to work with new p4 templates

PR #331, opened 2018-07-29
update autoqc.pm to use products to construct qc jobs

dozy:
I wonder if this code would be simpler if the parent check object used the moniker role to create file names. Then we would only had to supply qc_in.

### DIFF
--- a/lib/npg_pipeline/function/autoqc.pm
+++ b/lib/npg_pipeline/function/autoqc.pm
@@ -3,6 +3,7 @@ package npg_pipeline::function::autoqc;
 use Moose;
 use namespace::autoclean;
 use Readonly;
+use List::MoreUtils qw{any};
 use File::Spec;
 use Class::Load qw{load_class};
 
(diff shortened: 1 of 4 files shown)


lib/npg_pipeline/function/autoqc.pm
```diff
@@ -177,62 +183,80 @@ sub _create_definition_object {
...
     $c .= $self->is_paired_read() ? q[ --is_paired_read] : q[ --no-is_paired_read];
   } elsif ($check eq q[qX_yield] && $self->platform_HiSeq) {
     $c .= q[ --platform_is_hiseq];
   }
 
-  my $qc_out = defined $tag_index ? $self->lane_qc_path($position) : $self->qc_path();
-  $qc_out or $self->logcroak('Failed to get qc_out directory');
-  my $qc_in  = defined $tag_index ? $self->lane_archive_path($position) : $self->archive_path();
+  ###########
+  # set qc_in
```
mgcam: If we give files explicitly, we shoudl not need qc_in. Minor changes to parent check object might be needed.

lib/npg_pipeline/function/autoqc.pm
```diff
@@ -177,62 +183,80 @@ sub _create_definition_object {
...
-    $qc_in .= q[/.npg_cache_10000]
+  if(any { /$check/ } qw( gc_fraction qX_yield)) {
+    $c .= qq[ --input_files=$fqc1_filepath --input_files=$fqc2_filepath]
+  }
+  elsif(any { /$check/ } qw( insert_size ref_match sequence_error )) {
+    $c .= qq[ --input_files=$fq1_filepath --input_files=$fq2_filepath];
+  }
+  elsif(any { /$check/ } qw( adapter genotype verify_bam_id )) {
+    $c .= qq{ --input_files=$bamfile_path} # note: single bam file 
+  }
```
mgcam: Adapter might have to be treated separately since it's running parallel to seq_alignment and uses unaligned bam files. Does this mean that it will run on unmerged files? This will break assumptions I made in qc_store. We need to discuss this.

lib/npg_pipeline/function/autoqc.pm
```diff
@@ -88,20 +89,22 @@ sub create {
...
-    push @definitions, $self->_create_definition($h, $is_multiplexed_lane);
-
-    if ( $self->is_indexed && $is_multiplexed_lane) {
-      foreach my $tag_index (@{$self->get_tag_index_list($p)}) {
-        my %th = %{$h};
-        $th{'tag_index'} = $tag_index;
-        push @definitions, $self->_create_definition(\%th, $is_multiplexed_lane);
-      }
-    }
+  for my $dp (@{$self->products->{lanes}}) {
```
mgcam: Lane products might also be in data_products. Would it be better to create a list of unique data products and iterate over it? Otherwise, if we are not careful, some jobs might be created twice.

PR #335, opened 2018-07-31
S2merge autoqc

dozy:
change per-lane seqchksum comparison jobs to one per-run job
update spatial_filter to use products (construct complete input_files values)
update cluster_count to use products
correctly specify location for _t fastqcheck files in stage1


srl147:
tests to be fixed

PR #337, opened 2018-08-02
Pass qc_in to upstream tags QC

dkj:
so that it can find tag metrics QC for the lane from this analysis rather than looking up qc from disk for the run.


dkj:
wrong argument

PR #346, opened 2018-08-04
convert cluster count check into a run-level check

dozy:


lib/npg_pipeline/function/cluster_count.pm
```diff
@@ -150,6 +156,18 @@ sub run_cluster_count_check {
     return 1;
 }
 
+has 'bfs_paths' => ( isa        => 'ArrayRef',
+                     is         => 'ro',
+                     required   => 0,
+                     default  => sub {return [];}, # is sub necessary?
```
dkj: Generally yes for `default` as it is global....

PR #348, opened 2018-08-06
New tracking

jmtcsngr:


t/util.pm
```diff
@@ -83,9 +83,24 @@ sub create_multiplex_analysis {
   $self->create_analysis($args);
   my $analysis_runfolder_path = $self->temp_directory() . $ANALYSIS_RUNFOLDER_PATH;
   my $recalibrated_path = $self->temp_directory() . $RECALIBRATED_PATH;
-  `rm $analysis_runfolder_path/Recipe_GA2_37Cycle_PE_v6.1.xml`;
-  `cp t/data/Recipes/Recipe_GA2-PEM_MP_2x76Cycle+8_v7.7.xml $analysis_runfolder_path/`;
-
+  # 76 cycles 8 lanes Multiplex
+
```
mgcam: There changes are confusing. The RunInfo created here is not for an indexed run - this might be the reason is_indexed flag had to be set explicitly. I sugest to remove these RunInfo documents, all of them and then add the correct ones in thests that will fail (if any)

PR #357, opened 2018-08-10
Added s3_archiver pipeline function

keithj:


dkj:
There's a lot of files in this commit - but lots have been zeroed so presume that's intentional...

keithj:
I created a facsimile of a NovaSeq runfolder by traversing a real example to gather information about file naming and directory structure to use as test data. All the files are empty, except where necessary (e.g. an md5 file may contain the md5 of an empty file, rather than nothing).

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+      next;
+    }
+
+    my $job_name = sprintf q{%s_%d_%d}, $ARCHIVE_EXECUTABLE, $id_run, $i;
+    my $base_url = $archive_config->{$study_id};
+    $self->debug(sprintf "Using base URL '%s' for study %s",
+                 $base_url, $study_id);
+
+    my @aws_args = ('--cli-connect-timeout', '300',
+                    '--acl', 'bucket-owner-full-control');
```
dkj: Not necessary now but we could add `--profile <some profile picked up per study s3 config>` if such a profile exists in the study's s3 config...

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
+package npg_pipeline::function::s3_archiver;
+
```
mgcam: The new functions have to be registered in npg_pipeline::pluggable::registry, which provides mapping from names in input function graph to modules and methods

mgcam:
What about shifting this to npg_pipeline::runfolder_scaffold where other 
directories are defined and created?

On 10/08/2018 17:19, dkj wrote:
> Would like to shift this to somewhere within the analysis hierarchy - 
> archive is perhaps too far so either the top level of analysis 
> (bam_basecalls) or just above archive (no_cal).




-- 

lib/npg_pipeline/pluggable/registry.pm
```diff
@@ -48,7 +48,9 @@ Readonly::Hash my %REGISTRY => (
 
   'bam_cluster_counter_check'=> {'cluster_count' => 'create'},
   'seqchksum_comparator'     => {'seqchksum_comparator' => 'create'},
-                               );
+  'archive_to_s3'            => {'s3_archiver' => 'create'},
```
mgcam: Do we want to add the new functions to data/config_files/function_list_post_qc_review.json? Or are we goign to run them manually to start with?

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  return \@definitions;
+}
+
+sub _read_config {
+  my ($self) = @_;
+
+  my $config = {};
+  try {
```
mgcam: It would be safer to have a fatal error if we fail to read the config file. The deployment might go wrong, the file might not be where expected.

lib/npg_pipeline/pluggable/registry.pm
dkj: Best if they go straight into the post_qc_review pipeline (but that can be in either this PR or one soon after) - I'll be using this PR to test archival...

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+      my $file_url   = "$base_url/$sample/$filename";
+
+      my $command = sprintf q{%s s3 cp %s %s %s},
+        $ARCHIVE_EXECUTABLE, join(q{ }, @aws_args), $file_path, $file_url;
+
+      $self->info(sprintf q{S3 archiving %s in study %s to %s},
+                  $file_path, $study_id, $file_url);
+
+      push @definitions,
+      npg_pipeline::function::definition->new
```
dkj: Is there something extra which needs to be added here when returning `@definitions` with > 1 member? The jobs generated have the npg_pipeline log wrapper around them (needed for our big analysis jobs as wr only captures the beginning and end of logs) but they will overwrite each other.

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+  my ($self, $product) = @_;
+
+  $product or $self->logconfess('A product argument is required');
+
+  my @expected_files;
+
+  my $dir_path = catdir($self->archive_path(), $product->dir_path());
+  my @extensions = qw{cram cram.md5 cram.crai
+                      seqchksum sha512primesums512.seqchksum
+                      F0x900.stats F0xB00.stats F0xF04_target.stats
```
dkj: These should have the suffixes `_F0x900.stats` `_F0xB00.stats` and `_F0xF04_target.stats` (as we have the convention of using the `_` to indicate some sort of filtering/subsetting of the data), but bundled with these other suffixes it will look for `.F0x900.stats` `.F0xB00.stats` `.F0xF04_target.stats`.

lib/npg_pipeline/function/s3_archiver.pm
mgcam: 'composition' attribute should be copied from the product object to the definition object. wr job identifies and LSF array indexes are derived from composition objects. And, @dkj is correct here, output log file names

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+      $self->info(sprintf q{S3 archiving %s in study %s to bucket URL %s},
+                  $product->file_name_root, $study_id,
+                  $archive_config->{$study_id});
+    } else {
+      $self->info(sprintf q{Skipping S3 archiving %s in study %s},
+                  $product->file_name_root, $study_id);
+      next;
+    }
+
+    my $job_name = sprintf q{%s_%d_%d}, $ARCHIVE_EXECUTABLE, $id_run, $i;
```
mgcam: There is no need to index job names, for LSF it might be even wrong. It might be benefitial to use a more descriptive job name, say, copy_data2aws or aws_archival

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+      $self->info(sprintf q{S3 archiving %s in study %s to %s},
+                  $file_path, $study_id, $file_url);
+
+      push @definitions,
+      npg_pipeline::function::definition->new
+        ('created_by' => __PACKAGE__,
+         'created_on' => $self->timestamp(),
+         'identifier' => $id_run,
+         'job_name'   => $job_name,
+         'command'    => $command);
```
mgcam: Might be good to set memory as well if known.

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+    my $base_url = $archive_config->{$study_id};
+    $self->debug(sprintf q{Using base URL '%s' for study %s},
+                 $base_url, $study_id);
+
+    my @aws_args = qw{--cli-connect-timeout 300
+                      --acl bucket-owner-full-control};
+
+    foreach my $file_path ($self->expected_files($product)) {
+      my $filename   = basename($file_path);
+      my $file_url   = "$base_url/$sample/$filename";
```
mgcam: Shall we go for an extra level after the sample name in case we ever have to send multiple sets of files for the same sample or the names are not unique? Could be either the base name for the composition or composition digest.

lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -0,0 +1,265 @@
...
+
+__END__
+
+=head1 NAME
+
+npg_pipeline::function::s3_archiver
+
+=head1 SYNOPSIS
+
+  my $obj = npg_pipeline::function::s3_archiver
```
mgcam: npg_pipeline::function::s3_archiver->new ?

lib/npg_pipeline/function/s3_archiver.pm
keithj: Done

lib/npg_pipeline/function/s3_archiver.pm
keithj: Done

lib/npg_pipeline/function/s3_archiver.pm
keithj: Done

lib/npg_pipeline/function/s3_archiver.pm
keithj: Done

lib/npg_pipeline/function/s3_archiver.pm
keithj: Done

lib/npg_pipeline/function/product_delivery_notifier.pm
```diff
@@ -0,0 +1,335 @@
...
+       'excluded'   => 1);
+  }
+
+  return \@definitions;
+}
+
+sub _build_message_dir {
+  my ($self) = @_;
+
+  return catdir($self->archive_path(), q[..], q[..], $EVENT_MESSAGE_DIRNAME);
```
mgcam: More robust to use $self->analysis_path.

lib/npg_pipeline/function/product_delivery_notifier.pm
```diff
@@ -0,0 +1,335 @@
...
+    next if not $self->is_for_s3_release($product);
+    next if not $self->is_for_s3_release_notification($product);
+
+    my $msg_file = $product->file_path($self->message_dir, ext => 'msg.json');
+    $self->_write_message_file($msg_file, $self->make_message($product));
+
+    my $job_name = sprintf q{%s_%d_%d}, $SEND_MESSAGE_SCRIPT, $id_run, $i;
+    my $command =
+      sprintf q{%s --host %s --port %d --vhost %s --exchange %s},
+      $SEND_MESSAGE_SCRIPT, $self->message_host(), $self->message_port(),
```
mgcam: Not clear how host, port, etc get set.

lib/npg_pipeline/product/release.pm
```diff
@@ -0,0 +1,293 @@
...
+
+  my $rpt  = $product->rpt_list();
+  my $name = $product->file_name_root();
+  my $outcomes = npg_qc::mqc::outcomes->new(qc_schema => $self->qc_schema);
+  if ($outcomes->get_library_outcome($rpt)) {
+    $self->info("Product $name, $rpt is for release (passed manual QC)");
+    return 1;
+  }
+
+  $self->info("Product $name, $rpt is NOT for release (failed manual QC)");
```
mgcam: False value returned by get_library_outcome() does not mean 'fail' in manual qc. Just means that the product cannot be released

lib/npg_pipeline/product/release.pm
```diff
@@ -0,0 +1,293 @@
...
+
+sub _find_study_config {
+  my ($self, $product) = @_;
+
+  my $rpt      = $product->rpt_list();
+  my $name     = $product->file_name_root();
+  my $study_id = $product->lims->study_id();
+
+  $study_id or
+    $self->logconfess("Failed to get a study_id for product $name, $rpt");
```
mgcam: tag zero object might not have study_id defined

Build.PL
```diff
@@ -52,13 +53,16 @@ my $builder = $class->new(
           'MooseX::AttributeCloner'      => 0.17,
           'MooseX::Getopt'               => 0,
           'MooseX::StrictConstructor'    => 0,
+          'Net::AMQP::RabbitMQ'          => 0,
```
mgcam: What are prerequisites of Net::AMQP::RabbitMQ for deployment purposes?

dkj: Think we'll need the `librabbitmq-dev` and `uuid-dev` packages installed.

lib/npg_pipeline/function/product_delivery_notifier.pm
```diff
@@ -0,0 +1,335 @@
...
+    next if not $self->has_qc_for_release($product);
+    next if not $self->is_for_s3_release($product);
+    next if not $self->is_for_s3_release_notification($product);
+
+    my $msg_file = $product->file_path($self->message_dir, ext => 'msg.json');
+    $self->_write_message_file($msg_file, $self->make_message($product));
+
+    my $job_name = sprintf q{%s_%d_%d}, $SEND_MESSAGE_SCRIPT, $id_run, $i;
+    my $command =
+      sprintf q{%s --host %s --port %d --vhost %s --exchange %s},
```
dkj: Missing commit? Do these options need to be replaced with a `-c ~/.npg/psd_production_events.conf`?

PR #368, opened 2018-09-03
fixes to cluster_count and seqchksum_comparator checks

dozy:
Now that they are run-level checks and an analysis may use a subset of lanes,
explicitly specify lanes when creating the script commands for cluster_count
and seqchksum checks.

Remove the unneeded composition attribute when creating the command
definition (no more dummy lane position keys in the command file).

Update tests.


t/20-function-cluster_count.t
```diff
@@ -75,11 +69,9 @@ cp 't/data/run_params/runParameters.miseq.xml',  "$analysis_runfolder_path/runPa
   is ($values->{'default'}, 1, 'the queue is set to default for the definition');
   
   TODO: { local $TODO = 'currently returning one position - review';
```
mgcam: remove TODO ?

PR #369, opened 2018-09-10
Fixed a bug where existing message files were not checked

keithj:
Fixed a bug where existing message files were not checked for before
creating new ones.

Added a new feature where a new message file will be created if
tracking information changed since it was created.


lib/npg_pipeline/function/product_delivery_notifier.pm
```diff
@@ -154,7 +155,20 @@ sub create {
...
 
     my $msg_file = $product->file_path($self->message_dir, ext => 'msg.json');
-    $self->_write_message_file($msg_file, $self->make_message($product));
+    if (-e $msg_file) {
+      $self->info("Message file '$msg_file' exists");
+      if ($self->_message_content_changed($msg_file,
+                                          $self->make_message($product))) {
+        $self->info("Message content in file '$msg_file' has ",
+                    'changed. Writing new contents');
+        $self->_write_message_file($msg_file, $self->make_message($product));
```
sb10: Could calling make_message() twice be more expensive than storing its result in a variable? If so, just call it once?

lib/npg_pipeline/function/product_delivery_notifier.pm
```diff
@@ -233,6 +256,37 @@ sub _write_message_file {
...
+
+  my $old_message = $self->_read_message_file($file_path);
+  $self->debug('Comparing old message ', pp($old_message),
+               ' with new message ',     pp($new_message));
+
+  my $changed = 0;
+  # These metadata are sufficient to detect a LIMS change.
+  foreach my $attr (qw{customer_name sample_id
+                       sample_name sample_supplier_name
+                       file_path file_md5 rpt_list}) {
```
sb10: Could this list be defined elsewhere as a global "constant", with the specific meaning of "sufficient to detect a LIMS change", for this and potentially other functions to use?

PR #370, opened 2018-09-13
change components_as_products method of npg_pipeline::product to

dozy:
return a list with one item when there is only one component in
 the composition (instead of an empty list)


mgcam:
Travis CI tests fail to load Perl dependencies. Locally the branch merges OK and passes tests.

PR #377, opened 2018-10-11
tileviz index file with links to lane-level tileviz reports

mgcam:


mgcam:
See change to tileviz tracking page link in https://github.com/wtsi-npg/npg_tracking/pull/485

PR #378, opened 2018-10-26
archival daemon - limit number of simultaneously archived NovaSeq runs

mgcam:


lib/npg_pipeline/daemon/archival.pm
```diff
@@ -28,30 +20,31 @@ sub build_pipeline_script_name {
 sub run {
   my $self = shift;
 
+  my $submitted = 0;
   foreach my $run ($self->runs_with_status($ARCHIVAL_PENDING)) {
+    (not $submitted) or last;
```
sb10: The double negative can cause confusion and made me wonder if this was really intended. Since it is, how about:

```suggestion
    $submitted && last;
```

mgcam: Thank you, done.

PR #384, opened 2018-11-20
wr executor: interpret job's queue attr value

mgcam:
Specifically, request specific flavour for the new p4_stage1 queue
Also propagates ref repository root env var to wr jobs

### DIFF
--- a/Changes
+++ b/Changes
@@ -20,10 +20,14 @@ LIST OF CHANGES
    npg_pipeline::runfolder_scaffold role
  - minor changes for bcfstats qc check
  - executor type (lsf or wr) can be specified in the configuration file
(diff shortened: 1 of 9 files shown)


mgcam:
On 20/11/2018 12:50, Sendu Bala wrote:
>
> *@sb10* commented on this pull request.
>
> Is there no way to explicitly test that we you add a p4stage1 job, 
> that in wr the cloud_flavor setting is associated with it?
>
There is, but it's tricky to set up in the unit test for the executor 
itself. Easier to put this test into high-level pipeline tests which wiz 
through everything, but I am reluctant. So far the wr executor test 
coverage was not great. I'd like to try this out for real on OpenStack 
and then come back to the test.




-- 

PR #389, opened 2018-12-03
require 6GB memory for bcfstats

mgcam:


mgcam:
On 03/12/2018 12:59, Sendu Bala wrote:
>
> *@sb10* commented on this pull request.
>
> It would be good if these $MEMORY_* variables had better names that 
> tied to their usage. Why is it that bcfstats needs the same memory as 
> ref_match and pulldown_metrics? Coincidence?
>
Yes
>
> Just picking an existing $MEMORY_ variable?
>
Yes. It is expected that this will be reverted after a bit of bcfstats 
code optimisation




-- 

lib/npg_pipeline/function/autoqc.pm
```diff
@@ -183,14 +184,17 @@ sub _create_definition_object {
     $ref->{'command_preexec'} = $self->repos_pre_exec_string();
   }
 
+  ##no critic (ControlStructures::ProhibitCascadingIfElse)
```
sb10: Reimplement using hash at top, to avoid turning off critic

mgcam: Done

PR #391, opened 2018-12-03
Requirements groups4wr

mgcam:


lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -467,6 +468,7 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
     if(exists $methods_to_aligners{$p4_param_vals->{alignment_method}}) {
       $aligner = $methods_to_aligners{$aligner};
     }
+    $ref->{'req_group_hint'} = join q[_], 'gen', $aligner;
```
sb10: In all cases, should req_group_hint be appended with information regarding if a split was done? Or any other significant 'if' that happens after this block?

mgcam:
No longer needed

PR #397, opened 2018-12-19
follow change in unset_tag method signature and simplify the code

mgcam:
Follows from and depends on https://github.com/wtsi-npg/npg_tracking/pull/494


mgcam:
superseeded by https://github.com/wtsi-npg/npg_seq_pipeline/pull/400

PR #400, opened 2019-01-07
run_is_deletable reimplementation

mgcam:
Requirements:
1. the script can deal with merged data,
2. absence of data in iRODS does not present a problem (ie a list of files in iRODS is not used as a master list of extected files)
3. is easily extendable to accommodate additional product destinations (s3)

Limitations:
1. Requires presence of JSON files for product compositions and their subsets (if any)

Depends on https://github.com/wtsi-npg/npg_tracking/pull/494


mgcam:
Another pr is being prepared as an extension of this one. The code will not depend of the presence of composition JSON files, which are not available in early versions of new-style run folders. It would also be possible to make the code to work with old-style run folders.

PR #401, opened 2019-01-15
Destination aware irods publishing

mgcam:
Depends on https://github.com/wtsi-npg/npg_seq_pipeline/pull/400


mgcam:
Replace by https://github.com/wtsi-npg/npg_seq_pipeline/pull/402/files, which has clean history and does not depend on any other branch

PR #406, opened 2019-01-24
reduce a list of pipeline scripts arguments

mgcam:


dkj:
Don't we use `--tile_count` to do reduced data test runs? Or perhaps it is a red-herring and we have to use `--interactive` and alter the stage1 jobs manually in some way?

mgcam:
> Don't we use |--tile_count| to do reduced data test runs?
>
No. It's one of the long_info role attr which I hope we will be able to 
get rid of soon
>
> Or perhaps it is a red-herring and we have to use |--interactive| and 
> alter the stage1 jobs manually in some way?
>
Yes.




-- 

PR #408, opened 2019-01-25
run is deletable reimplementation

mgcam:
Reimplementation of 'run_id_deletable' script to make it work with multi-component entities and per-product archive directories. Ability to work with old-style runfolders is retained. The iRODS part of the validation process is aware of product destination. It is not yet aware of a per-product iRODS collection hierarchy, so is not yet suitable for NovaSeq runs.

### DIFF
--- a/lib/npg_pipeline/validation/autoqc.pm
+++ b/lib/npg_pipeline/validation/autoqc.pm
@@ -198,8 +198,9 @@ sub _results_exist {
 	}
       } else {
 
-        my $local_expected = $expected;
-        my $local_count    = $count;
+        my $local_expected     = $expected;
+        my $local_count        = $count;
+        my $local_compositions = $compositions;
 
         if ($check_name eq 'adapter') {
           # Adapter check for tag zero might or might not be present.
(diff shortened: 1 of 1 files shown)


bin/npg_run_is_deletable
```diff
@@ -1,278 +1,34 @@
...
-}
+$v->info(sprintf 'Running %s for run %i, runfolder %s',
+                 __FILE__, $v->id_run, $v->runfolder_path);
+
+my $deletable = $v->run();
+
+my $m = sprintf 'Run %i is %sdeletable.', $v->id_run, $deletable ? q[] : q[NOT ];
+$v->info($m);
+
+exit $deletable ? 0 : 1;
```
dkj:
```suggestion
exit ($deletable ? 0 : 1);
```
else precedence is such that  `exit $deleteable` is evaluated before the `?`

mgcam: (not $deletable) is logically correct, but returns an undefined valie in case $deletable is true. Is this acceptable for a script return value?

dkj: If think by the time it gets to the system call it will be converted to integers in the range 0 to 127.
```bash
$ perl -le 'exit not 1'; echo $?
0
$ perl -le 'exit not 0'; echo $?                                                                                                                                                                
1
```

mgcam: Thank you, done

PR #415, opened 2019-02-08
Set messages as persistent (delivery_mode 2) and added some comments.

kjsanger:


bin/npg_pipeline_notify_delivery
```diff
@@ -132,10 +132,13 @@ if (not $mq->is_connected()) {
 }
 $mq->channel_open($channel);
 
+# 'mandatory' ensures the message reached at least one queue
```
mgcam: This setting, when enabled, also makes the message returned to the sender in case it cannot be placed. I am not sure how we can capture this on the client side. We should certainly explote the return value of the recv function. For example, when timeout is reached without getting a reply, undefined is returned.

dkj: I cannot find info on this. Do we know behavior in this code changes and for what scenarios? I hoped this tweak only impacted message behaviour on RabbitMQ server (n.b. I think we only have to improve the situation from what we have - we can accept some imperfection at the moment...).

mgcam:
On 08/02/2019 11:59, dkj wrote:
> I cannot find info on this.
https://www.rabbitmq.com/reliability.html


-- 

mgcam:
On 08/02/2019 11:59, dkj wrote:
> I hoped this tweak only impacted message behaviour on RabbitMQ server
Might change return as well, but, since we are not checking returned 
info, nothing changes for us


-- 

mgcam:
On 08/02/2019 11:59, dkj wrote:
> I think we only have to improve the situation from what we have
Requesting persistence improves the situation, ie messages should 
survive most problems with the server


-- 

bin/npg_pipeline_notify_delivery
kjsanger: The `mandatory` setting is not a change - it was tested and present in the initial deployment. I just added this comment as a housekeeping improvement while I was making other changes nearby.

mgcam:
On 08/02/2019 12:32, Keith James wrote:

> The |mandatory| setting is not a change
Indeed. The change is an extra line after it. Apologies. Still, I 
believe an error on undefined return from $mq->recv($rv_timeout); would 
be an improvement. Not that I request a change



-- 

PR #417, opened 2019-02-08
function list for combined archival pipeline

mgcam:
QC data archival moved upfront. Archival of autoqc data validates autoqc JSON files which are then sent to s3 and iRODS archives. Running upload to qc database before ml warehouse upload ensures that the latter is able to load a complete data set (except the 'qc complete' status date) and does not access the file system when retrieving autoqc data.


mgcam:
>  how about replacing the existing post_qc_review pipeline
For seq farm I'd like to add a pre-requisite to LSF job (existence in outgoing) to prevent the job from occupying a slot needlessly for 20 mins and to remove this condition from the next two jobs.

PR #418, opened 2019-02-12
no_s3_archival flag

mgcam:


dkj:
> I was going to comment that it would be better not to have the negative flag (no pun intended), to avoid conversations like "Is archival on?" "Yes, no archical is not enabled". Then I checked the other flags and they're all of that form, so I think that horse has bolted... !

Agree - both  better and bolted...

PR #419, opened 2019-02-12
seq_alignment function: cache unsuccessfully retrieved references

mgcam:


lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -779,33 +772,50 @@ sub _ref {
   my $ref_name = $self->_do_gbs_plex_analysis ? $dplims->gbs_plex_name : $dplims->reference_genome();
 
   my $ref = $ref_name ? $self->_ref_cache->{$ref_name}->{$aligner} : undef;
-  if (!$ref) {
+  my $na_flag = 'NOT AVAILABLE';
```
dkj: Candidate for `Readonly::Scalar`?

mgcam: Thanks, done.

PR #428, opened 2019-03-12
images for function graphs

mgcam:


jmtcsngr:
I think there is a problem with one of the images. The file name is a dot-something and I can not open it as PNG

`._function_list_post_run_archived.json.png`

PR #433, opened 2019-03-20
set explicit umask for wr jobs

mgcam:
Patch to master to be released on OS ASAP.
Tested by manual edit of a file that is used in wr add -f command for wr manager running with a cloud config option. The jobs create output files that are writable for a group.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,9 @@
 LIST OF CHANGES
 ---------------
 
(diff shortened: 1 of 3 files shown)


mgcam:
> N.b. `umask 2` would work as well as `umask 0002` or `umask 002` I think

002 definitely does

PR #456, opened 2019-05-07
handle cluster_count values of zero or less - just use zero for subsa…

dozy:
just use zero for subsample fraction (i.e. don't divide by zero)


dozy:
extra tests not yet added

PR #460, opened 2019-06-10
Add facility to do LSF 1:1 job dependancies on array jobs

mp15:
Not quite finished yet but this is how I would do it for LSF


mp15:
With Marina's changes having fixed the perl critic and tests this is probably ready for testing.

PR #472, opened 2019-07-11
Create singleton to hold ref cache

mp15:
Take ref cache code from seq_alignment and move to stand alone class.


mp15:
Updated with a few tweaks to tidy up code before merge.

PR #475, opened 2019-07-12
Haplotype caller

mp15:
HaplotypeCaller initial working draft. There are probably a few things we want to change, but this works!


mp15:
Depends on #482

PR #490, opened 2019-08-05
Add MD5 to bucket upload command if available

dkj:


lib/npg_pipeline/function/s3_archiver.pm
```diff
@@ -134,6 +135,17 @@ sub _check_files {
   return;
 }
 
+sub _base64_encoded_md5_gsutil_arg {
+# if there is a corresponding .md5 file for the given path, use its contents to
+# add an MD5 header for the data being uploaded.
+  my ($self, $path) = @_;
+  $path .= q(.md5);
+  if (not -e $path){ return; }
```
mgcam: return an empty string to avoid warning
'perl -le 'use strict; use warnings; print join q[ ], 1,2,undef,3,4;'
Use of uninitialized value in join or string at -e line 1.
1 2  3 4'

dkj: I thought I might hit that too, but
```
dj3@sf2-farm-srv1:~$ perl -le 'use strict; use warnings; print join q[ ], 1,2,undef,3,4;'                                                                                                                          
Use of uninitialized value in join or string at -e line 1.
1 2  3 4
dj3@sf2-farm-srv1:~$ perl -le 'use strict; use warnings; sub foo{return;};print join q[ ], 1,2,foo(),3,4;'
1 2 3 4
```

dkj: I think it's the LIST context and it's an empty list....

dkj: Good in so far as I don't get an extra space character in the commands I generate.

mgcam: OK. Magic Perl ways

mgcam:
`
gsutil -h Content-MD5:xp6LyjGWpl86Al9kfX7AJQ== cp gtest.txt gs://some_test/
Copying file://gtest.txt [Content-Type=text/plain]...
BadRequestException: 400 Provided MD5 hash "xp6LyjGWpl86Al9kfX7AJQ==" doesn't match calculated MD5 hash "xp6LyjGWpl86Al9kfX7tJQ==".
echo $?
1
`

PR #491, opened 2019-08-06
BQSR table calculation function

mp15:
Built on #484 and #475 (both now merged).


lib/npg_pipeline/function/bqsr_calc.pm
```diff
@@ -0,0 +1,170 @@
...
+  my @definitions = ();
+
+
+  foreach my $product (@products) {
+    $self->is_release_data($product) or next;
+
+    my $dir_path = $product->path($self->archive_path());
+
+    my $input_path = $product->file_path($dir_path, ext => 'cram');
+    my $output_path = $product->file_path($dir_path, ext => 'table');
```
mgcam: If we use .bqsr.table or _bqsr.table or something similar, it would be easier to see straight away what kind of file this is

lib/npg_pipeline/function/bqsr_calc.pm
```diff
@@ -0,0 +1,170 @@
...
+                 @{$self->products->{data_products}};
+
+  if (scalar @products == 0) {
+    $self->debug('no BQSR enabled data products, skipping');
+  }
+  my @definitions = ();
+
+
+  foreach my $product (@products) {
+    $self->is_release_data($product) or next;
```
mgcam: Might be better to move this filter as an extra grep in lines 47-48.

dkj:
Should we use the `npg_common::roles::software_location` role for `gatk` executable?

lib/npg_pipeline/function/bqsr_calc.pm
dkj: n.b. only keen on the `_` instead of a `.` to separate filename root from suffix when the `_something` indicates a filtering of the data (like with the various stats files we produce)

lib/npg_pipeline/function/bqsr_calc.pm
mp15: done

lib/npg_pipeline/function/bqsr_calc.pm
mp15: done

mp15:
Revised to make changes for review.

t/20-function-bqsr_calc.t
```diff
@@ -0,0 +1,159 @@
...
+use Cwd qw/cwd abs_path/;
+use Perl6::Slurp;
+use File::Copy;
+use Log::Log4perl qw/:levels/;
+use JSON;
+use Cwd;
+use List::Util qw/first/;
+
+use Moose::Util qw(apply_all_roles);
+
```
mgcam: It should be possible to remove lines 19, 52-56, 127-129 from the test without any adverse effect.

t/20-function-bqsr_calc.t
```diff
@@ -0,0 +1,159 @@
...
+  $rna_gen->create_product_level();
+
+  my $da = $rna_gen->create();
+
+  ok ($da && @{$da} == 12, sprintf("array of 12 definitions is returned, got %d", scalar@{$da}));
+
+  my $command = qq{$gatk_exec BaseRecalibrator -O $dir/novaseq/180709_A00538_0010_BH3FCMDRXX/Data/Intensities/BAM_basecalls_20180805-013153/no_cal/archive/plex4/26291#4.bqsr_table -I $dir/novaseq/180709_A00538_0010_BH3FCMDRXX/Data/Intensities/BAM_basecalls_20180805-013153/no_cal/archive/plex4/26291#4.cram -R $fasta_dir/GRCh38_15_plus_hs38d1.fa --known-sites $annot_dir/dbsnp_138.hg38.vcf.gz --known-sites $annot_dir/Mills_and_1000G_gold_standard.indels.hg38.vcf.gz};
+
+  my $mem = 2000;
+  my $d = _find($da, 1, 4);
```
mgcam: my $d = $da->[3];
should also work since the definitions follow the sort order of the products (12 merged products ordered by tag index); lines 92-107 can then be removed

mp15: done

t/20-function-bqsr_calc.t
mp15: done

PR #505, opened 2019-08-28
configurable data deletion delay

mgcam:
Goes well with https://github.com/wtsi-npg/npg_seq_pipeline/pull/508 and https://github.com/wtsi-npg/npg_tracking/pull/534

### DIFF
--- a/t/10-validation.t
+++ b/t/10-validation.t
@@ -1,6 +1,6 @@
 use strict;
 use warnings;
-use Test::More tests => 4;
+use Test::More tests => 5;
 use Test::Exception;
 use Test::Warn;
 use Test::Trap qw/ :warn /;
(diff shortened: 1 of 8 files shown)


t/10-validation.t
```diff
@@ -70,14 +70,58 @@ subtest 'create object' => sub {
   ok ($v->use_cram, 'cram files are used by default');
   is ($v->file_extension, 'cram', 'default file extension is cram');
   is ($v->index_file_extension, 'crai', 'default index file extension is crai');
-  is ($v->min_keep_days, 14, '14 days after qc complete data to ve retained');
```
dkj: Should this go back in to test the default? Or be removed completely?

mgcam: This test cannot be run in the context of 'create object' sub-test since the build method uses a list of products, which cannot  be computed with limited input this object gets. Changing this test to show that the value can be overwritten.

PR #510, opened 2019-08-29
no sort and no markdup for spike tag

dozy:
amend splice directive in seq_alignment for spike tag (e.g. #888) to skip sort and markdup steps


dozy:
Examples of graphs produced with this pull request - spike tag (#888), and tag#0 and tag#1 to show they still have their sort and markdup

run_30590_1#888
run_30590_1#0
run_30590_1#1

dozy:
Updated to leave in the coordinate sort. Leaving it out introduced unexpected complications.

PR #515, opened 2019-09-09
Add necessary code for BWA MEM 2

mp15:


mp15:
This may require a small modification to p4 (straight copy of bwa_mem p4 method to bwa_mem2)

mgcam:
It's not clear from the code changes how use of bwamem2 is triggered. Is it possible to set up a test that clarifies this?

mgcam:
Thanks for the test. We need to add the new test samplesheet t/data/miseq/samplesheet_24135_bwa_mem2.csv to the MANIFEST

PR #519, opened 2019-09-17
LSF chunked jobindex

mp15:
Prevent us from using our fancy job id system when making chunked jobs.


mp15:
Didn't work will try again

PR #520, opened 2019-09-18
Move chunk directory to no_archive path

mp15:
We don't want to archive chunked files for HaplotypeCaller so move them to different point in runfolder. Depends on wtsi-npg/npg_tracking#536


mgcam:
I think we will need to create no_archive path and per-product directories on npg_pipeline::runfolder_scaffold

mp15:
Updated to include tests and runfolder_scaffold.

PR #521, opened 2019-09-18
code simplification: use existing data product objects

dkj:


dkj:
Can this work? Am I missing something?

PR #524, opened 2019-09-30
delete intermediate data

dozy:
create list of intermediate files for later deletion in seq_alignment and add new remove_intermediate_data function to post_qc_review pipeline to actually do the deletions.

### DIFF
--- a/data/config_files/function_list_post_qc_review.json
+++ b/data/config_files/function_list_post_qc_review.json
@@ -69,6 +69,11 @@
          {
             "relation" : "dependsOn",
             "source" : "run_run_archived",
+            "target" : "remove_intermediate_data"
+         },
+         {
+            "relation" : "dependsOn",
+            "source" : "remove_intermediate_data",
             "target" : "run_qc_complete"
          },
          {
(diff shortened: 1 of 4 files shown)


mgcam:
My personal preference is not to have interdependencies between modules in npg_pipeline::function and not to have in these modules any code that is not needed for generation of the function definition.

mp15:
The alternative could be to put the intermediate files in the no_archive directory, going to add a few tweaks to the PR but it’s nearly finished.

dkj:
Chatted with @dozy @mgcam and @jmtcsngr : this calculation for explicit files to delete and passing that state around seems much more expensive (in complexity, and in breaking and introducing conventions) than I'd anticipated. So, having learned from this I'd recommend (simpler approach of) an intermediate files deletion pipeline function which launches a `rm -v <recal_dir>/*.cram` job which we run in the archival pipeline before `qc complete` (so avoid the directory move issue).

dkj:
Simplified method was easier to review!
Tested on R&D run - looked good.

PR #525, opened 2019-10-01
receipt generation

mgcam:


dkj:
Tried and used successfully on real data.

PR #534, opened 2019-10-23
add -f to rm command

dozy:
add -f to rm command when removing intermediate cram files to avoid error and job failure when no intermediate crams are present


mgcam:
Something like this to avoid -f flag?
eval 'ls dodo'; MYERR=$?; if [ $MYERR = 0 ]; then (echo 'remove files' && rm dodo); else echo 'nothing to do'; fi;

PR #538, opened 2019-11-19
Move BAM to no_archive

mp15:
Move intermediate BAM files to the no_archive path as we do not add them to IRODS and this also makes them easier to delete if we need disk space urgently.


mgcam:
https://github.com/wtsi-npg/npg_seq_pipeline/blob/d717d7f9ced5985c1e542dd0224f67cab7a0a2c2/lib/npg_pipeline/function/autoqc.pm#L218 also needs changing

PR #549, opened 2020-01-13
make test CRAM files compliant with samtools v.1.10.0

mgcam:


ces:
Tested with samtools 1.9 & 1.10 on seqfarm.

PR #554, opened 2020-02-14
add function to run interop autoqc check

mgcam:


mgcam:
repeat of https://github.com/wtsi-npg/npg_seq_pipeline/pull/546

PR #560, opened 2020-03-13
remove old wh loader function from analysis

mgcam:


dkj:
[Image](https://github.com/mgcam/npg_seq_pipeline/blob/57bb204a6384fd64f9e69a011e5b8cc4681c7cf2/data/config_files/function_list_central_qc_run.json.png) still has old warehouse loader in?

PR #563, opened 2020-03-19
when markdup_method is "none", add skip_markdup_metrics flag to bam_f…

dozy:
…lagstats qc command


dozy:
depends on https://github.com/wtsi-npg/p4/pull/273 (though only when the markdup_method is "none")

PR #567, opened 2020-03-24
Groups limits for wr executor

mgcam:
This pr also removes array_cpu_limit for LSF jobs as a separate commit. This was used to avoid contention on lane collections creation, which, as @kjsanger has shown, is not a problem

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 LIST OF CHANGES
 ---------------
 
+ - ability to apply limits to wr groups of jobs and a limit for
+   all iRODS jobs
  - function creating definitions for autoqc jobs - when evaluating
    whether the autoqc check should be run:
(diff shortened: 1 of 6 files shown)


dkj:
I've two thoughts on this:
- I'd like the option to just specify the `limit_grps` with a name but without a limit (I envisage this limit is something we would want to ramp up and down e.g. for an hour whilst iRODS is down, but forcing a number in here will reset it in the wr manager whenever a job is submitted)
- I wonder if we could simplify this and `p4stage1_queue` pipeline config to a generic "mixin" to the JSON pushed to `wr add` i.e. for any pipeline function we specify json/data-structure which we merge into the JSON job info sent to `wr add`

I don't think either of these should block this PR.

PR #568, opened 2020-03-30
a new function for running a third party analysis within Nextflow

mgcam:
Depends on https://github.com/wtsi-npg/npg_tracking/pull/573


lib/npg_pipeline/function/stage2pp.pm
```diff
@@ -0,0 +1,135 @@
...
+    push @definitions, npg_pipeline::function::definition->new($attrs);
+  }
+
+  foreach my $product (@products) {
+
+    my $in_dir_path  = $product->stage1_out_path($self->no_archive_path());
+    my $out_dir_path = $product->path($self->archive_path());
+
+    my $command = join q[ ], $self->nextflow_cmd(), 'run',
+                             'wtsi-team112/ncov2019-artic-nf',
```
jmtcsngr: This probably needs to be a whole path, either in lustre or part of DD.

lib/npg_pipeline/runfolder_scaffold.pm
```diff
@@ -45,6 +48,22 @@ sub create_product_level {
...
+    #
+    # A link is created without an error even if its target does not exit.
+    # The target will be created during the p4stage1 step.
+    #
+    my $ext = 'cram';
+    foreach my $p ( @{$self->products->{'data_products'}}) {
+      my $link   = $p->file_path($p->stage1_out_path($self->no_archive_path()), ext => $ext);
+      if (!-l $link) { # a link might exist if we are re-using the analysis directory
+        my $target = $p->file_path($self->recalibrated_path(), ext => $ext);
+        symlink $target, $link or croak "Failed to create a symlink $link";
```
dkj, line 64: Could we consider using  `abs2rel` here so the symlinks aren't broken when the runfolder is moved (e.g. when moving to outgoing),

PR #574, opened 2020-04-14
Add COG-UK metadata upload client

kjsanger:
POSTs sequencing library and run metadata to the COG-UK metadata API
endpoint described at
https://docs.covid19.climb.ac.uk/upload-instructions.

### DIFF
--- a/lib/npg_pipeline/product/heron/upload/metadata_client.pm
+++ b/lib/npg_pipeline/product/heron/upload/metadata_client.pm
@@ -35,7 +35,6 @@ has 'api_uri' =>
   (isa           => 'URI',
    is            => 'ro',
    required      => 1,
-   default       => sub { return URI->new('https://localhost') },
    documentation => 'COG-UK API endpoint URI to receive sequencing metadata',);
 
 =head2 send_library_metadata
--- a/lib/npg_pipeline/product/heron/upload/metadata_client.pm
+++ b/lib/npg_pipeline/product/heron/upload/metadata_client.pm
@@ -1,0 +1,284 @@
+package npg_pipeline::product::heron::upload::metadata_client;
+
+use strict;
+use warnings;
+
+use Data::Dump qw(pp);
+use JSON;
+use LWP::UserAgent;
+use Moose;
+use MooseX::StrictConstructor;
+use URI;
+
+with qw[
+    WTSI::DNAP::Utilities::Loggable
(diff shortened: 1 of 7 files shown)


dkj:
I don't think we should force supplying of artic primers version as:
- it's not compulsory at the far end
- it may not be per run, but per sample-library/plex
- we may use non-artic primers

It  would be a good thing for us to fill in this value when we know it - we can parse it from `primer_panel` in `iseq_flowcell`

kjsanger:
I'm not sure what you mean with respect to defaults. None of the classes have defaults for attributes - all must be supplied. The classes do validate those attributes based on the what the metadat API expects.  None of the classes make their own DB queries. All queries and assignment of values happens in one place (the script).

mgcam:
My mistake about db queries, confused by very similar code. As far as defaults are concerned, one of the examples:
npg_pipeline::product::heron::upload::library line 14, package-level variable
`our $NEB_ULTRA = q(NEB Ultra II);`
in  bin/npg_heron_metadata_upload, line 160, assignment of default
`$sequencing_kit           ||= 'NEB Ultra II';`
To minimise repetition and guard against typos in further edits, could be
`$sequencing_kit           ||= $npg_pipeline::product::heron::upload::library::NEB_ULTRA;`
My personal preference is for a separate package with constants, which are private variable that are exported on demand or for a config file. However, it's just a personal preference and too much work right now.

lib/npg_pipeline/product/heron/upload/metadata_client.pm
```diff
@@ -0,0 +1,288 @@
...
+    (isa           => 'Str',
+     is            => 'ro',
+     required      => 1,
+     documentation => 'The registered access token for the COG-UK API account',);
+
+has 'api_uri' =>
+  (isa           => 'URI',
+   is            => 'ro',
+   required      => 1,
+   default       => sub { return URI->new('https://localhost') },
```
mgcam, line 38: Would it be better not to have a default here?

mgcam:
MANIFEST file update is required

kjsanger:
Added issue #578 for the above

lib/npg_pipeline/product/heron/upload/metadata_client.pm
kjsanger: Default removed

kjsanger:
MANIFEST updated

PR #576, opened 2020-04-20
Make providing an ARTIC primers version optional on the CLI

kjsanger:
Requires #574


kjsanger:
Rebased on `devel` which has the MANIFEST updates

PR #583, opened 2020-04-29
Added 'showback' to end of run to store runtime statistics.

jenniferliddle:


jenniferliddle:
> The only purpose of classes in the npg_pipeline::function namespace is to generate definitions of pipeline functions, they are not meant to contain any code that is executed by the jobs.

I don't know what they're _meant_ to contain, but I can tell you that they _do_ contain code. The precedent is in cluster_count.pm, seqchksum_comparator.pm, s3_archiver.pm, just to pick some at random. I don't recall which one I chose as a model.

jenniferliddle:
> not clear why executor type should be passed to the pipeline while it's available in the configuration file

It may not be in the configuration file, or it may be in the configuration file but wrong because it's been overridden on the command line.

jenniferliddle:
> qc_path is no longer in use

I can confirm it definitely is in use, because I have used it. If you mean it's been deprecated, what is the recommended replacement?

jenniferliddle:
Replaced by #611

PR #584, opened 2020-04-29
Archiver4climb

mgcam:
Built on the top of https://github.com/wtsi-npg/npg_seq_pipeline/pull/581

Two pipeline functions are created, 'pp_archiver' and 'pp_archiver_manifest', the former one will be included in the function graph for archival, the latter one just generates a manifest and is intended to be used manually. An existing manifest specified by an env variable NPG_MANIFEST4PP_PATH can be used  by 'pp_archiver'. If a file at the path specified by NPG_MANIFEST4PP_PATH does not exist, it is created. Criteria for products to manifest  - the  pp is flagged for external archival (ie we will not flag a configuration for the R&D study) and uqc pass.

The new function creates a job which will call a as yet not written script, which will take the manifest path as input, stage a directory with hard links and send the data off. Should also check that each of the samples is not  at CLIMB site already

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -64,6 +64,7 @@ my $builder = $class->new(
           'Readonly'                     => 1.03,
           'strict'                       => 1.03,
           'Sys::Filesystem::MountPoint'  => 0,
+          'Text::CSV'                    => 0,
           'Try::Tiny'                    => 0,
           'UUID'                         => 0,
           'URI'                          => 0,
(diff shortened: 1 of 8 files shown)


dkj:
I fear the logic for what to go in the manifest will change e.g. criteria to choose which of a sample duplicate to use - maybe push it to the upload script which could also check the remote for duplicates. But good to get this functionality in now...

PR #590, opened 2020-05-28
script for data upload to CLIMB

mgcam:


bin/npg_upload2climb
```diff
@@ -0,0 +1,274 @@
...
+my $manifest4upload_file = $manifest_file;
+if ( not $manifest4upload_file =~ s/$MANIFEST_EXTENTION\Z/${MANIFEST_SUFFIX}${MANIFEST_EXTENTION}/xms) {
+  $manifest4upload_file .= $MANIFEST_SUFFIX . $MANIFEST_EXTENTION;
+}
+
+my $to_upload = [];
+my $e;
+try {
+  if (-e $manifest4upload_file) {
+    $logger->info("Found existing manifest for upload $manifest4upload_file");
```
dkj, line 63: Minor variant: What about _always_ running the code in this block to read the upload manifest, and moving the upload manifest generation above (conditional on file not existing, and not filling in the `$to_upload` directely) - so there is only one code path for filling in `$to_upload`.

bin/npg_upload2climb
```diff
@@ -0,0 +1,274 @@
...
+  $archives = _scaffold_staging_archive($to_upload);
+} catch {
+  $e = $_;
+  $logger->error("Error scaffolding staging archive: $e");
+  $logger->info('Exiting');
+};
+
+$e and exit 1;
+
+if ($dry_run) {
```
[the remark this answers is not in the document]
mgcam: Implemented, thanks

bin/npg_upload2climb
mgcam: Implemented; was leaning to this myself

dkj:
There are a few tweaks we could consider:
- don't insist on ssh related arguments but just pass those unused by the script through to ssh (this could, for example, allow us to specify a single config file and no other ssh options are required)
- perhaps we should alert the user that dry-run does actually create a `climbupload` file which will be used in future non-`--dry_run` runs...

PR #599, opened 2020-07-02
append autoqc generic result generation at the end of pp

mgcam:
also see https://github.com/wtsi-npg/npg_qc/pull/696


dkj:
@mgcam notes that we need to ensure we're capturing the pipeline version...

PR #604, opened 2020-08-05
Stage2pp refactored to create common methods

mgcam:
in anticipation of adding a second pp


lib/npg_pipeline/function/stage2pp.pm
```diff
@@ -72,8 +94,10 @@ sub create {
     }
   }
 
+  @definitions = grep { $_ } @definitions;
+
   if (@definitions) {
-    (@definitions == @{$self->_output_dirs}) or $self->logcroak(
+    (@definitions <= @{$self->_output_dirs}) or $self->logcroak(
       sprintf 'Number of definitions %i and output directories %i do not match',
```
jmtcsngr, line 101: I think this error message needs to be updated to match the condition checked. 'Less or equal than' vs 'match'.

jmtcsngr (changes requested):

PR #607, opened 2020-08-10
Archive PP products to iRODS

dkj:
Build on #606 but:
- cope with no sample supplier name
- use `sample_supplier_name` instead of `supplier_sample_name` to match existing iRODS metadata (evnen though columns in warehouse `iseq_*_product_metrics` tables don't match)


dkj:
changes adopted by #606

PR #611, opened 2020-09-07
Another attempt to add 'showback' function

jenniferliddle:


dkj:
Thanks for separating out code into separate modules by functionality - I think we can push this a bit further: e.g. put thenew modules in a new namespace `npg_pipeline::showback` (or `npg_pipeline::executor::showback` ) rather than in `npg_pipeline::executor` space (though I acknowledge some of this is per executor - so see the + side ). We could move the `process_run` into a `npg_pipeline/showback.pm`
I've not yet looked at code - only structure...

PR #612, opened 2020-09-16
Fix no repeat sample data regression and stop syncing timestamps

dkj:


mgcam (approved):
Tested the amended ssh command live (listing sample directories) and subsequent glob to get the sample name. Noted that the --size-only option is used with rsync. But, in fact, we never upload over the existing directory.

PR #618, opened 2020-11-30
added npg_climb2mlwh

jenniferliddle:


dkj (changes requested):
We need a few changes:
- move job functionality out of `function` namespace, maybe to existing `heron` space
- alter `pp_archiver` function to create a job which runs the new script as well as running the existing the upload script
- purge both unnecessary `use` from module code and the commented out code

t/20-function-climb2mlwh.t
```diff
@@ -0,0 +1,78 @@
...
+
+my $dir = tempdir( CLEANUP => 0);
+
+use_ok('npg_pipeline::function::climb2mlwh');
+
+# Create dummy command file
+my $exec = join q[/], $dir, 'wsi-npg-ssh-restricted';
+open my $fh1, '>', $exec or die 'failed to open file for writing';
+print $fh1 qq[
+printf 'perl: warning: Setting locale failed.\\n'
```
dkj, line 22: The warnings come back on STDERR so should not be mixed with the real data returned I think,

lib/npg_pipeline/product/heron/upload/climb2mlwh.pm
```diff
@@ -0,0 +1,273 @@
...
+use strict;
+use warnings;
+use Moose;
+use namespace::autoclean;
+use MooseX::Getopt::Meta::Attribute::Trait::NoGetopt;
+use MooseX::StrictConstructor;
+use English qw(-no_match_vars);
+use Carp;
+use Readonly;
+use Getopt::Long;
```
dkj, line 12:
```suggestion
```
superfluous with `MooseX::Getopt` ?

lib/npg_pipeline/product/heron/upload/climb2mlwh.pm
```diff
@@ -0,0 +1,273 @@
+package npg_pipeline::product::heron::upload::climb2mlwh;
+
+use strict;
+use warnings;
```
dkj, line 4:
```suggestion
```
Included with `use Moose` no?

bin/npg_climb2mlwh
```diff
@@ -0,0 +1,145 @@
...
+#use English qw(-no_match_vars);
+use Carp;
+#use Readonly;
+#use Getopt::Long;
+#use Pod::Usage;
+#use Log::Log4perl qw(:easy);
+#use Try::Tiny;
+#use DateTime;
+#use WTSI::DNAP::Warehouse::Schema;
+#use npg_tracking::glossary::composition::factory::rpt_list;
```
dkj, line 16:
```suggestion
use Carp;
```
purge commented code...

dkj (changes requested):
Can `Pod::Usage` be removed too?

dkj (changes requested):
Looking really close now. Possible refinements:
- add your new test file to the MANIFEST
- move the `easy_init` to the script (that is _mostly_ not done in our modules)
- `->logcroak` in module (it is unusual to `exit` in a module I think, and rather than `croak` as you're using Log4Perl)
- since we're using Log4Perl here, we can (should?) use it instead of `print` (and then use `--verbose` to set log level of INFO perhaps?)
- info currently printed in the dry run section could go into an `info` log message always, dry-run skipping should proably `->warn(q(Dry-run so not updating DB))` or similar
- `MooseX::Getopt::Meta::Attribute::Trait::NoGetopt` unused so purge?

PR #626, opened 2021-01-15
remove hard dependency of tests on the presence of NPG tools

mgcam:
allow tests that do not require iRODS operations proceed if the baton executable are not available

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,10 @@
 LIST OF CHANGES
 ---------------
 
+ - use a mock execubable in one of the tests to allow all tests in
+   this package to run successfully if NPG tools stack is not available
+   (excluding tests that are run against a test iRODS server)
+
 release 58.1.0
  - added npg_climb2mlwh to update warehouse from uploaded
    climb data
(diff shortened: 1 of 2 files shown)


nerdstrike:
I feel somehow that this ought to be the responsibility of WTSI::NPG::iRODs to operate in a test mode, rather than all other consumers having to fake its env (bearing in mind that I don't know how WTSI::NPG::iRODs operates).

dkj:
How about only using this mocking if `TOOLS_INSTALLED` is _not_ set. We have used that elsewhere before:
```
t/20-function-seqchksum_comparator.t:##     skip 'no tools', 2 if ((not $ENV{TOOLS_INSTALLED}) and (system(q(which bamseqchksum)) or system(q(which scramble))));
```
(I'm not sure why this example is block commented out :disappointed:  )

kjsanger:
> I feel somehow that this ought to be the responsibility of WTSI::NPG::iRODs to operate in a test mode, rather than all other consumers having to fake its env (bearing in mind that I don't know how WTSI::NPG::iRODs operates).

The env is only required if you actually try to connect. Creating an instance of `WTSI::NPG::iRODs` is what causes a search for a compatible client program i.e. `baton` (which still doesn't cause connection) - that's the tool in question which needs to be present to avoid an error at that point.

PR #627, opened 2021-01-15
Duplex-Seq

srl147:


mgcam:
For p4 stage 1 flag _is_duplexseq is set for the whole run if at least one library type is Duplex-Seq. Is this correct? Don't we need to approach this per lane?

srl147:
> For p4 stage 1 flag _is_duplexseq is set for the whole run if at least one library type is Duplex-Seq. Is this correct? Don't we need to approach this per lane?

Yes, it should be per lane. I've fixed this and updated the PR

PR #629, opened 2021-01-18
iRODS connections to be opened on-demand for validation

kjsanger:
When iRODS is unavailable, skip tests that construct an
npg_pipeline::validation::irods object (which requires an iRODS handle
and hence an iRODS client).

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 LIST OF CHANGES
 ---------------
 
+ - iRODS connections to be opened on-demand for validation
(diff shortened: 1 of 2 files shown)


dkj:
@mgcam @kjsanger Is this an alternative to #626 ?

kjsanger:
> @mgcam @kjsanger Is this an alternative to #626 ?

Yes, offered because @mgcam wasn't 100% happy with making a fake executable. I don't have a strong opinion either way, though.

mgcam (approved):

PR #633, opened 2021-02-01
refactor pipeline daemon code

mgcam:


lib/npg_pipeline/daemon/archival.pm
```diff
@@ -122,6 +121,8 @@ Invokes the archival pipeline for runs with a status 'archival pending'.
```
jmtcsngr, line 121: I think `sleep_time_between_runs` can also be removed from POD.

lib/npg_pipeline/daemon.pm
```diff
@@ -172,43 +108,10 @@ sub staging_host_match {
...
-
-  my @studies = ();
-  if (!$lims->{'qc_run'}) {
-    @studies = uniq map { $_->study_id } grep { !$_->is_control } @fcell_rows;
-    @studies = sort @studies;
-  }
-  $lims->{'studies'} = \@studies;
-
-  return $lims;
+  return {id => $batch_id};
```
jmtcsngr, line 114: I was thinking this method can be changed to `get_batch_id`. Then just return the id or logcroak. But the analysis daemon child depends on a hash to be returned. So it will need to be refactored there too with a few lines changed. So not sure how you feel about the whole chain of changes.

jmtcsngr (changes requested):

jmtcsngr (approved):

PR #634, opened 2021-02-02
Remove qc run flag

mgcam:


dkj:
I did use the `--lims_driver_type` back in 2016 - so it wasn't worthless adding it! Happy to purge now.

mgcam:
On 12/02/2021 10:02, dkj wrote:
>
> I did use the |--lims_driver_type| back in 2016 - so it wasn't 
> worthless adding it! Happy to purge now.
>
The |lims_driver_type| attribute was used in a number of unrelated 
places in the pipeline. Having it as a pipeline option gave an 
impression that it's possible to set this option in all cases, which is 
not true. It's not difficult to restore this option, but if we need to 
do this in future, I'd use a more specific name, for example, 
|lims_driver_type4samplesheet_generation|




-- 

PR #638, opened 2021-02-08
implementation for wr limit groups is simplified

mgcam:


dkj (approved):
Functionality change as desired. Tests pass in my hands.

PR #642, opened 2021-02-15
flexible mapping of wr group limits to jobs and a new limit for s3

mgcam:


jmtcsngr (commented):
I forgot if we are updating copyright legend for modules we touch. If so, we may need to update `wr.pm`

PR #643, opened 2021-02-15
do not limit the number of NovaSeq runs being archived

mgcam:
Depends on https://github.com/wtsi-npg/npg_seq_pipeline/pull/642


mgcam:
We will limit to four runs pushed to archival within an hour. A different pr will be created

PR #649, opened 2021-03-02
Handle tags that are too small

jenniferliddle:
For dual tags, remove common suffixes on both tags independently


lib/npg_pipeline/cache/barcodes.pm
```diff
@@ -19,6 +19,8 @@ Readonly::Scalar my $TAG_LIST_FILE_HEADER      => qq{barcode_sequence\tbarcode_n
 # I've extended these to 10 bases. For I5OPPOSITE final `GT` predicted from adapter documentation.
 Readonly::Scalar my $SPIKED_PHIX_TAG2        => q{TCTTTCCCTA};
```
dkj, line 20:
```suggestion
Readonly::Scalar my $SPIKED_PHIX_TAG2        => q{TCTTTCCCTACAC};
```
Extra bases from @srl147 analysis.

lib/npg_pipeline/cache/barcodes.pm
```diff
@@ -230,6 +232,25 @@ sub _get_tag_length {
...
+  my ($self, $index, $index_lengths) = @_;
+  my @idx = split /-/smx, $index;
+  my @ilengths = @{$index_lengths};
+  # if there is only one index length set the second index length to 0
+  if (scalar @ilengths == 1) { push @ilengths, 0; }
+  # if there is no separator in the index set the second part of the index to empty
+  if (scalar @idx == 1) { push @idx, q{}; }
+  # add suffix if required
+  if (length $idx[0] < $ilengths[0]) { $idx[0] .= $TAG_SUFFIX; }
+  if ( (length $idx[1] < $ilengths[1]) && length $idx[1]) { $idx[1] .= $TAG2_SUFFIX; }
```
dkj, line 247: This will depend on whether the i5 is opposite or not, no? I.e. you'll need a `I5OPPOSITE_TAG2_SUFFIX` in that case...

srl147: I think I've already replied to this elsewhere. The problem is we don't know the sequence to pad the i5 with if it is opposite. I suggest we write code to check the direction but if the i5 is opposite we simply output an error. As soon as we have lane which generates the error we can work out what sequence to pad the tags with.

dkj:
Closing for now as re-evaluation of requirements made it clear a rewrite of key functionality is required to meet them and also have maintainable code.

PR #652, opened 2021-03-09
Refactor log linking code and add copying of product_release.yml

nerdstrike:
Puts a copy of the product_release.yml file into the QC output folder, e.g. `...../tmp/201005_MS7_35178_A_MS3046657-600V3/Data/Intensities/BAM_basecalls_20201009-124801/product_release_20210309-150724-554939295.yml`

### DIFF
--- a/Changes
+++ b/Changes
@@ -6,11 +6,13 @@ LIST OF CHANGES
  - An additional wr limit group - s3 - is configured.
  - The wait4path pipeline job does not have a log. To enable wr to
    recognise that these jobs are unique, echoing a random string
-   is added to the shell command. 
+   is added to the shell command.
  - Remove the limit on the number of NovaSeq runs being archived at
    the same time. Introduce a limit on the number of runs,
    irrespectively of the instrument type, which are moved to archival
    within the last hour.
+ - Persist product_release.yml file to data directory to preserve run
+   conditions.
 
 release 59.0.0
  - Following a decision to send data to CLIMB regardless of artic
@@ -47,7 +49,7 @@ release 58.3.0
 release 58.2.0
  - enhancement of code for interaction with the Majora/COG-UK API
  - added analysis for Duplex-Seq libraries
- 
+
 release 58.1.0
  - added npg_climb2mlwh to update warehouse from uploaded
    climb data
@@ -86,7 +88,7 @@ release 57.15.0
  - append autoqc generic result generation at the end of ncov2019_artic_nf
    portable pipeline
  - tests update following a change in the default behaviour of the
-   add_object method in WTSI::NPG::iRODS 
+   add_object method in WTSI::NPG::iRODS
 
 release 57.14.0
  - switch to sample control flag when determining eligibility for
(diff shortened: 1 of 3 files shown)


Changes
```diff
@@ -6,11 +6,13 @@ LIST OF CHANGES
  - An additional wr limit group - s3 - is configured.
  - The wait4path pipeline job does not have a log. To enable wr to
    recognise that these jobs are unique, echoing a random string
-   is added to the shell command. 
+   is added to the shell command.
  - Remove the limit on the number of NovaSeq runs being archived at
    the same time. Introduce a limit on the number of runs,
    irrespectively of the instrument type, which are moved to archival
    within the last hour.
+ - Persist product_release.yml file to data directory to preserve run
```
mgcam, line 14: Change 'to data directory' - 'to analysis directory'
to avoid confusion with the `Data` directory of the run folder?

lib/npg_pipeline/pluggable.pm
```diff
@@ -701,32 +702,55 @@ sub _run_spider {
...
 
-  (-e $self->analysis_path) or return;
+sub _save_product_conf_to_analysis_dir {
+  my ($self) = @_;
+  # First check for unusual file names:
+  my ($filename, $dirs, $suffix) = fileparse(
+    $self->product_conf_file_path,
+    qr/[.][^.]*/xms
+  );
+  if ($filename ne 'product_release') {
```
mgcam, line 721: <($filename !~ /product_release/)> ?
Otherwise `product_release-mg.yml` is copied to `product_release_product_release-mg_20210318-171328-425726136.yml`

lib/npg_pipeline/pluggable.pm
```diff
@@ -701,32 +702,55 @@ sub _run_spider {
...
+  my ($self, $source_file, $override_name) = @_;
+
+  my $analysis_path = $self->analysis_path;
+  (-e $analysis_path) or return;
+  my $target_file;
+  if ($override_name) {
+    $target_file = catfile($analysis_path, $override_name);
   } else {
-    $self->warn("Will not overwrite existing file $link");
+    $target_file = catfile($analysis_path, $source_file);
```
mgcam, line 746: This path through the code is not used, neither it is tested. I believe it will create an incorrect target file path since the analysis path is concatenated with the path of the source file.

lib/npg_pipeline/pluggable.pm
```diff
@@ -701,32 +702,55 @@ sub _run_spider {
...
+  }
+  my $state = copy $source_file, $target_file;
+  if (!$state) {
+    $self->error("Failed to make a copy of $source_file at $target_file");
+  } else {
+    $self->info("Created copy of $source_file into $target_file");
```
mgcam, line 752: Might be good to move this info line to above line 748 (changing to 'Creating a copy of...) so that this line is available in the file before we copy it. Was not an issue when the file was linked.

lib/npg_pipeline/pluggable.pm
nerdstrike: Turns out it is used earlier in the tests where no log_file_name has been set for the pluggable instance. You're right about the catfile being suspect.

PR #659, opened 2021-04-13
Removing old travis files from MANIFEST

FredDodd6:
Removing old travis files. 
Adding GitHub workflow file.


mgcam:
Is the branch from which the pr is made is up-to-date- with wtsi/devel? There is a conflict in MANIFEST

PR #662, opened 2021-05-04
detect Hi-C library types and set appropriate bwa mem flags (via p4 p…

dozy:
…arameters)


dozy:
This depends on https://github.com/wtsi-npg/p4/pull/287

PR #664, opened 2021-05-17
Observe gbs_pipeline enabled or not in seq_alignment

ces:
Also needs associated but non urgent tweaks in genotype and genotype_call qc checks (genotype_call only called from seq_alignment and genotype only run for human samples) to avoid issues in future.


t/data/release/config/seq_alignment/product_release.yml
```diff
@@ -1,23 +1,35 @@
 ---
 default:
   markdup_method: "biobambam"
-
+  gbs_pipeline: &gbs_pipeline_accessible
+    allowed: true
 study:
   - study_id: "3573"
     markdup_method: "samtools"
+    gbs_pipeline: *gbs_pipeline_accessible
```
mgcam, line 9: We only need to specify this entry if it is different from the default.

PR #667, opened 2021-06-11
Configure status change jobs to save statuses to a database.

mgcam:
Saving to the database is disabled when either the 'local' or
'no_db_status_update' options are set to true.

Depends on https://github.com/wtsi-npg/npg_tracking/pull/615

### DIFF
--- a/lib/npg_pipeline/base/options.pm
+++ b/lib/npg_pipeline/base/options.pm
@@ -104,6 +104,22 @@ has q{no_warehouse_update} => (
   documentation => q{Switches off updating the NPG warehouse.},
 );
 
+=head2 no_db_status_update
+
+Switches off creating a database run and lane status record during the status
+update job execution. False by default. Is automatically set to true when the
+C<local> flag is set to true. 
+
+=cut
+
+has q{no_db_status_update} => (
+  isa           => q{Bool},
+  is            => q{ro},
+  lazy          => 1,
+  builder       => '_default_to_local',
+  documentation => q{Switches off updating database run and lane statuses.},
+);
+
 =head2 local
 
 Sets the default for no_irods_archival, no_warehouse_update and
--- a/lib/npg_pipeline/base/options.pm
+++ b/lib/npg_pipeline/base/options.pm
@@ -117,13 +117,13 @@ has q{no_db_status_update} => (
   is            => q{ro},
   lazy          => 1,
   builder       => '_default_to_local',
-  documentation => q{Switches off updating database run and lane statuses.},
+  documentation => q{Switches off run and lane status updates to the database},
 );
 
 =head2 local
 
-Sets the default for no_irods_archival, no_warehouse_update and
-no_summary_link to true.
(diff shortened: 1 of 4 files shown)


lib/npg_pipeline/function/status.pm
```diff
@@ -59,6 +59,10 @@ sub _command {
     }
   }
 
+  if (not $self->no_db_status_update) {
```
nerdstrike, line 62: Never been a fan of double negatives in code, but this will pass

lib/npg_pipeline/function/status.pm
```diff
@@ -77,6 +81,9 @@ npg_pipeline::function::status
 =head1 DESCRIPTION
 
 Launches a job for saving run and lane statuses to a file.
+If saving of statuses to the database is not disabled (C<--no_db_status_update>
```
nerdstrike, line 84: Too complex/passive. Legibility would be easier with something like "The job will also save statuses to the tracking database unless --no_db_status_update or --local options are used"

lib/npg_pipeline/base/options.pm
```diff
@@ -104,6 +104,22 @@ has q{no_warehouse_update} => (
...
+C<local> flag is set to true. 
+
+=cut
+
+has q{no_db_status_update} => (
+  isa           => q{Bool},
+  is            => q{ro},
+  lazy          => 1,
+  builder       => '_default_to_local',
+  documentation => q{Switches off updating database run and lane statuses.},
```
nerdstrike, line 120: Perhaps "Switches off run and lane status updates to the database" ?

t/20-function-status.t
```diff
@@ -59,7 +61,7 @@ my %init = (
     'job_name is correct');
   is ($d->command,
     'npg_status2file --id_run 1234 --status "run archived" --dir_out '
```
nerdstrike, line 63: Will you change this invocation first, or will that come after the other PR is merged?

nerdstrike (changes requested):
Looks ok, I'd like a little improvement to the comments.

t/20-function-status.t
mgcam: After the tracking pr is merged

nerdstrike: That's done, so you're free to update here too

nerdstrike (approved):

nerdstrike:
Final thought before merge. Have any relevant SOP's been updated, and does anyone need telling at next DD release?

mgcam:
> Have any relevant SOP's been updated
The description of the new option appears in help. Using `--local` option is recommended in SOP when testing the pipeline, which will automatically prevent saving statuses to the  db. `--local` flag ensures that the directory with the new analysis is not visible to the production daemons, including the status daemon. Thus, before this change` one of the consequences of using the `--local` flag was the new statuses **not** appearing in the database. I'll add this explanation to the Changes file.
> and does anyone need telling at next DD release?
We will definitely include this into release notes for the new DD.

PR #672, opened 2021-07-02
Adding tests and module code for preparing barcodes.

FredDodd6:
Final state from Fred Dodd.
Foundation for barcode rework.


dkj:
We need to add the case of a zero length i5 expected sequence (where the "padding" is different...)

dkj:
Incorporated in #685. Thanks!

PR #677, opened 2021-07-28
Support new no_auto, no_auto_archive, and no_auto_analysis tags

### LINKED ISSUE #645: no_auto , no_auto_analysis, no_auto_archive tags
to prevent runs going to analysis or archival

jenniferliddle:
Support new no_auto, no_auto_archive, and no_auto_analysis tags for daemons
Fixes #645


t/data/dbic_fixtures/300-Run.yml
```diff
@@ -45,6 +45,19 @@
...
+  batch_id: 55
+  expected_cycle_count: 66
+  flowcell_id: dummy_flowcell1
+  id_instrument: 48
+  id_instrument_format: 7
+  id_run: 1235
+  id_run_pair: 0
+  is_paired: 0
+  priority: 1
+  folder_name: 123456_IL2_1234_A_80B5TABXX
```
jmtcsngr, line 58: Collisions with folder names is a bad idea because we still use folder names to get run ids in some cases. It may not cause problems now but may cause unexpected results for other people updating tests in the future.

```suggestion
  folder_name: 123457_IL2_1235_A_81B5TABXX 
```

t/50-npg_pipeline-daemon-analysis.t
```diff
@@ -33,6 +33,8 @@ local $ENV{PATH} = join q[:], $temp_directory, $current_dir.'/t/bin', $ENV{PATH}
 
 my $dbic_util = t::dbic_util->new();
 my $schema = $dbic_util->test_schema();
+my $test_run_1235 = $schema->resultset(q[Run])->find(1235);
+$test_run_1235->update_run_status('analysis pending', 'pipeline',);
```
jmtcsngr, line 37: As I get it from the fixtures this will only be testing `no_auto_analysis` and not testing `no_auto`. I think is best if we test both paths.

If I understand correctly the idea will be to check that by introducing the 2 new `no auto` tags for analysis but not setting it for any other runs, you expect everything to stay the same. Then this run 1235 is used to test a run with the `no_auto_analysis` tag will not go into analysis. Which I guess is kind of fair but it was difficult to understand.

If you keep things as they are now you may also need to change some other lines se line 94 `ok(scalar @test_runs >= 3, 'at least three runs are analysis pending');` which I guess should become `>= 4`. Line 95 should include 1235. And so on in the file. 

So maybe easier to create a new subtest and do run + tag 52 produces no change in status, run + tag 53 produces no change in status and run without tag gets status change. Otherwise you may end having to re-check half of the old test code which seems as an overkill for your change.

t/50-npg_pipeline-daemon-archival.t
```diff
@@ -171,7 +171,7 @@ subtest 'limiting number of runs being archived' => sub {
   ($rst->description eq $status) or die 'inconsistent test data';
   $rst->update({'date' => DateTime->now()->subtract(hours => 2)});
   $s = $runner->run();
-  is ($s, 1, 'run submitted since the number of runs recently submitted ' .
+  is ($s, 0, 'run submitted since the number of runs recently submitted ' .
```
jmtcsngr, line 174: The message needs to be updated if things kept as currently are. But similar to the comment in tests for analysis. Here you are modifying a subtest which is meant to check we can limit the number of simultaneous runs in archival. So I think I understand why you made the change here, but anyone else not aware of the context will find it confusing.

jmtcsngr (changes requested):
The tests for this code are (before your change) a mix of styles and are very complex. I think I understand your changes, but I struggled to get the idea. I made some comments about trying to keep your tests isolated in new subtests. What do you think?

jenniferliddle:
Now tested by running npg_pipeline_analysis_runner and npg_pipeline_archival_runner against the dev database. My feeling is that this is now ready for squashing and merging.

PR #678, opened 2021-08-02
add missing CPAN dependencies

dkj:


dkj:
Spurious install leading me astray...

PR #684, opened 2021-08-12
functions to return run and product collections

mgcam:
This functionality was requested by @jidur to be used for reheading code in npg_seq_melt.


jidur (commented):
Tested with npg_seq_melt::util::change_header unit test and with NovaSeq run.

PR #686, opened 2021-08-18
Updated pulldown metrics can use CRAM files

nerdstrike:
Depends on [npg_qc change](https://github.com/wtsi-npg/npg_qc/pull/775)

Pulldown metrics will be performed on CRAM rather than BAM in future


mgcam:
Tested by running the analysis with the new version of the pulldown metrics check.

PR #696, opened 2021-09-14
Do not install a separate conda client.

mgcam:


dkj (approved):
Given Conda preinstalled in image...

PR #705, opened 2022-02-04
Defined deletion checks for pp products

mgcam:


lib/npg_pipeline/validation.pm
```diff
@@ -670,13 +670,43 @@ sub _irods_seq_deletable {
...
+  ######
+  # A simplified procedure, which will stop and return 0 (not deletable)
+  # as soon as something goes wrong or the first incorrectly archived
+  # file is found.
+  # No checks are done for products for which no pp was run, these
+  # products are considered deletable in the context of this method.
+  #
+  foreach my $p (@{$self->product_entities}) {
+    # get study conf for a product
+    # if 'irods_pp' > 'filters' > 'include' sectin exists
```
jmtcsngr, line 696: I like the idea of using product release for picking the files. But you were right about it being very specific. I see the pp version is there. I remember it was possible to use a specific product release to manually execute the pipeline. Guess we can use something like that to delete older versions and use the current with a shorter staging_deletion_delay for more recent runs. Guess we will have a similar issue every time we switch versions but we can always address it manually until we have a general solution.

lib/npg_pipeline/validation.pm
```diff
@@ -670,13 +670,43 @@ sub _irods_seq_deletable {
...
+    #   - apply include/exclude filter to the pp product staging archive
+    #     to get a list of files
+    #   - too many or too few files - error
```
jmtcsngr, line 699: I don't know how we can check for few files than expected. We have a pattern `v(1.4)\d+\b\S+trim\S+/\S+cram` which I never understood what it is matching.

mgcam:
Implemented in https://github.com/wtsi-npg/npg_seq_pipeline/pull/707

PR #707, opened 2022-02-09
iRODS deletable check is extended to pp data

mgcam:


t/10-validation.t
```diff
@@ -292,4 +292,44 @@ subtest 'flagged as not deletable' => sub {
...
+
+  delete $ref->{per_product_staging_archive};
+  $ref->{per_product_archive} = 1;
+  $v = npg_pipeline::validation->new($ref);
+  ok ($v->per_product_staging_archive, 'per product archive is true');
+  ok ($v->per_product_archive, 'per product iRODS is true');
+  is ($v->irods_destination_collection, '/seq/illumina/runs/8/8747',
+    'iRODS collection');
+
+  $ref->{irods_destination_collection} = '/seq-dev/8747';
```
jmtcsngr, line 328: A `/seq-dev` with `$ref->{per_product_archive} = 1;` should map to `/seq-dev/illumina/runs/8/8747` right?

t/10-validation.t
```diff
@@ -292,4 +292,44 @@ subtest 'flagged as not deletable' => sub {
...
+  my $rfh = _create_test_runfolder_8747();
+  my $ref = {
+    id_run => 8747,
+    runfolder_path      => $rfh->{'runfolder_path'},
+    analysis_path       => $rfh->{'analysis_path'},
+    archive_path        => $rfh->{'archive_path'},
+    qc_schema           => $qc_schema,
+    npg_tracking_schema => $tracking_schema
+  };
+  mkdir join q[/], $rfh->{'runfolder_path'}, 'npg_do_not_delete';
```
jmtcsngr, line 307: Not sure this was used. I was expecting to see a validation returning false because of the folder. But don't think it is there.

t/10-validation.t
mgcam: In this test the collection is set via constructor, meaning that the value remains as set regardless of the value of per-product flag. The test demonstrates a possibility of overwriting the logic implemented in the code.

t/10-validation.t
mgcam: This test was not changed in this pr

mgcam: Removed, thanks

PR #713, opened 2022-02-22
Add irods_locations directory to runfolder

mksanger:
Directory is populated with a json file per product during archival.  These files can be loaded into the seq_product_irods_locations mlwh table using the npg_irods_locations2ml_warehouse script in npg_ml_warehouse


mgcam (approved):
The new feature is currently enabled for both archival and analysis pipeline. We'll run with this for now. Should not cause any problems.

PR #717, opened 2022-03-04
remove product delivery notification system

dkj:


dkj:
FYI @kjsanger - dropping functionality you added for UKBV. Think it has served its purpose and is ok to drop now - right?

mgcam (approved):
dry run for both analysis and archival pipeline is OK

PR #719, opened 2022-03-07
Add irods location loading step

mksanger:


mksanger:
This is dependent on https://github.com/wtsi-npg/npg_seq_pipeline/pull/713, and should not be merged until it is confirmed that the JSON files are being created in the right place following a new DD containing that pr.

mgcam (approved):
The new feature is currently enabled for both archival and analysis pipeline. We'll run with this for now. Should not cause any problems.

mgcam (commented):
No comment at this point

mgcam (changes requested):
No specific changes requested at this point, just trying to withdraw the approval

PR #721, opened 2022-03-16
reverse logic for SamHaplotag --revcomp flag 

dozy:
it should be used when is_i5_opposite attribute is false

### DIFF
--- a/lib/npg_pipeline/function/seq_alignment.pm
+++ b/lib/npg_pipeline/function/seq_alignment.pm
@@ -314,6 +314,9 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
   my $is_haplotag_lib = $l->library_type && ($l->library_type =~ /Haplotagging/smx);
   if($is_haplotag_lib) {
     $p4_param_vals->{haplotag_processing} = q[on];
+    # the samhapotag tool was developed with data from an i5 "rev comp workflow",
+    # its --revcomp flag allows it to work with "standard workflow" i5, as would be
+    # obtained from MiSeq or NovaSeq V1.0 reagents, hence the "not"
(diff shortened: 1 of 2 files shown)


lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -314,7 +314,7 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
   my $is_haplotag_lib = $l->library_type && ($l->library_type =~ /Haplotagging/smx);
   if($is_haplotag_lib) {
     $p4_param_vals->{haplotag_processing} = q[on];
-    if($self->is_i5opposite) {
+    if(not $self->is_i5opposite) {
```
dkj, line 317:
```suggestion
    # the samhapotag tool was developed with data from a i5 "rev comp workflow"
    # its revcomp flag allows it to work with "standard workflow" i5, as would be obtained
    # from MiSeq or NovaSeq V1.0 reagents
    if(not $self->is_i5opposite) {
```
Worth adding comment?

lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -314,7 +314,10 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
   my $is_haplotag_lib = $l->library_type && ($l->library_type =~ /Haplotagging/smx);
   if($is_haplotag_lib) {
     $p4_param_vals->{haplotag_processing} = q[on];
-    if($self->is_i5opposite) {
+    # the samhapotag tool was developed with data from an i5 "rev comp workflow",
```
dkj, line 317:
```suggestion
    # the samhaplotag tool was developed with data from an i5 "rev comp workflow",
```

PR #727, opened 2022-06-20
Removed provisions for loading the old warehouse

mgcam:


dkj:
drop `update_warehouse_post_qc_complete` at same time?

PR #728, opened 2022-06-20
Run params loader

mgcam:
Built on top of https://github.com/wtsi-npg/npg_seq_pipeline/pull/727

Depends on https://github.com/wtsi-npg/npg_ml_warehouse/blob/devel/bin/npg_run_params2mlwarehouse being released.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,11 @@
 LIST OF CHANGES
 ---------------
+
+ - Removed provisions for loading the old warehouse, the function
+   for loading the old warehouse was removed some time ago.
+
 release 61.2.0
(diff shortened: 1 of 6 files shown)


mgcam:
On 22/06/2022 15:45, dkj wrote:
>
> Looks fine - but remind me of logic for limiting to post qc complete, 
> please. Could we run it for all ml_warehouse loading jobs?
>
We can, I just decided that since that job is really only to get the qc 
complete timestamp, we do not need to make it more complex than it is. 
Also, I was too lazy to deal with a change of runfolder path from 
analysis to outgoing.
>
> Message ID: ***@***.***>
>


-- 

PR #735, opened 2022-07-21
single-end runs to use markdup_method:samtools for stage2 analysis

dozy:
Force p4 stage2 parameter markdup_method to value "samtools" when run is single-end (and target alignment is on).


dozy:
Note that when do_target alignment is false, all markdup nodes are spliced out, so no need to adjust markdup_method

lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -426,7 +426,7 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
     $p4_param_vals->{reference_dict} = $self->_ref($dp, q(picard)) . q(.dict);
     $p4_param_vals->{reference_genome_fasta} = $self->_ref($dp, q(fasta));
     if($self->p4s2_aligner_intfile) { $p4_param_vals->{align_intfile_opt} = 1; }
-    $p4_param_vals->{markdup_method} = $self->markdup_method($dp);
+    $p4_param_vals->{markdup_method} = ($self->is_paired_read ? $self->markdup_method($dp) : q[samtools]);
```
dkj, line 429: Wasn't expecting this dependent on target alignment - though that may be appropriate.

We were considering something like:
```
if (not  $self->is_paired_read and not $self->markdup_method($dp) ) {
  # override markdup method for single read runs as we experience occasional hangs using default biobambam
  $p4_param_vals->{markdup_method} = q[samtools];
}
```

dozy: You are right - this will not work since it doesn't change phix or other subsets when there is no target alignment.

But I think it should be just:
if (not  $self->is_paired_read) {
  # override default markdup method for single read runs as we experience  
  # occasional hangs using default (biobambam)
  $p4_param_vals->{markdup_method} = q[samtools];
  $self->info(q[Overriding markdup method for single-end, always use samtools]);
}

In other words, use markdup_method:samtools for single-end even if product_release.yml says otherwise. Do you agree?

dkj: I think that is ok: my use of `and not $self->markdup_method($dp)` was to _try_ to make it conditional on the use of the default.

t/20-function-seq_alignment.t
```diff
@@ -1755,4 +1756,106 @@ is_deeply($h, $expected, 'correct json file content for run 37416 lane 2 tag 1 p
...
+      conf_path         => 't/data/release/config/seq_alignment',
+      resource          => $default
+    )
+  } 'no error creating seq_alignment object';
+
+  apply_all_roles($hs_gen, 'npg_pipeline::runfolder_scaffold');
+  $hs_gen->create_product_level();
+
+  my $da = $hs_gen->generate('analysis_pipeline');
+warn q[da count: ], scalar @{$da};
```
dkj, line 1795:
```suggestion
```

t/20-function-seq_alignment.t
```diff
@@ -1755,4 +1756,106 @@ is_deeply($h, $expected, 'correct json file content for run 37416 lane 2 tag 1 p
...
+  my $d = _find($da, 1, 1);
+  isa_ok ($d, 'npg_pipeline::function::definition');
+
+  ## check json file for lane 1 tag 1
+  my $json_file = qq{$bc_path/45421_1#1_p4s2_pv_in.json};
+  ok (-e $json_file, 'json params file exists for run 45421 lane 1 tag 1');
+  my $h = from_json(slurp($json_file));
+
+warn q[h: ], Dumper($h);
+
```
dkj, line 1807:
```suggestion
```

t/data/release/config/seq_alignment/product_release.yml
```diff
@@ -32,4 +32,5 @@ study:
     markdup_method: "samtools"
     gbs_pipeline: &gbs_pipeline_inaccessible
       allowed: false
-
+  - study_id: "6931"
+    markdup_method: "biobambam"
```
dkj, line 36: Needed? Slightly surprised - but content...

dkj (approved):
I think the warns in the test should be removed - replace with `diag` if need be...

PR #750, opened 2022-10-28
Allow for human split for all data

mgcam:


Changes
```diff
@@ -1,6 +1,12 @@
 LIST OF CHANGES
 ---------------
 
+ - run deletion:
+   account for a potential human split for GBS runs and
+     lanes where a primer panel is set,
+   error message extended with info about a product when pp files are
+   missing on staging
```
jmtcsngr, line 8: I think this only applies to data retention so maybe worth explaing

```suggestion
   missing on staging when checking for data retention
```

PR #771, opened 2023-05-25
bwa mem2 overrides

dozy:
- add bwa_mem2 flag to options.pm to allow override of default bwa analyses at pipeline invocation
- update seq_alignment to recognise the bwa_mem2 flag and also default to bwa_mem2 for NovaSeqX platform
- add tests and test data


dkj (approved):
Code logic looks cleaner and understandable for the various combinations of method, executable, reference etc.

Not tried - content to merge on understanding we'll run full suite of DD test runs (setups) to test.

PR #775, opened 2023-06-02
remove constraint of p4s1 to specific flavor

jmtcsngr:
Wondering if it is best to remove the key `p4stage1_queue`. There is other test in the same `t/30-executor-wr.t` file which checks that if there is a value it is picked up. I think the test I changed was there just to make sure the key force the flavour we wanted. @dkj @mgcam


dkj (approved):
Have tried creating the new wr commands (without `"cloud_flavor":"ukb1.2xlarge",`) - looks fine.

PR #784, opened 2023-07-14
tweak to seq_alignment so GbS samples with no study ref do not fail

ces:


lib/npg_pipeline/function/seq_alignment.pm
```diff
@@ -429,7 +429,7 @@ sub _alignment_command { ## no critic (Subroutines::ProhibitExcessComplexity)
     $p4_param_vals->{markdup_method} = $do_gbs_plex ? q[none] : $self->markdup_method($dp);
     $p4_param_vals->{markdup_optical_distance_value} = ($uses_patterned_flowcell? $PFC_MARKDUP_OPT_DIST: $NON_PFC_MARKDUP_OPT_DIST);
 
-    if($p4_param_vals->{markdup_method} eq q[none]) {
+    if(!$do_gbs_plex && ($p4_param_vals->{markdup_method} eq q[none])) {
```
dkj, line 432: Is this right? Disabling mark dup is itself disabled for GbS?

dkj: Yes. 
> `$skip_target_markdup_metrics = 1;` is set for GbS line 489 (previously 488)
> The change was to handle the rare case where the GbS study has no study/sample ref but we are forcing target align and a GbS reference. So trying simply to avoid line 435 ;
`if(my $pcb=npg_pipeline::cache::reference->instance->get_primer_panel_bed_file($dp, $self->repository)) {`
As get_primer_panel_bed_file dies when there is no study/sample ref.

PR #786, opened 2023-07-18
per product publish doc and variable name fixup

dkj:


mgcam (approved):
Excellent. Thanks for editing the documentattion.

PR #791, opened 2023-07-28
switch off spatial filter for NovaSeqX

dozy:


dkj:
Can we add a test? There is a NovaSeqX folder in `t/` now...

dkj (approved):
Following discussion: tests to come later...

PR #797, opened 2023-08-29
Switch to Perlbrew to obtain multiple Perl versions

kjsanger:
Change some file names and labels to be consistent with other Perl CI


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+          cpanm --quiet --notest Module::Build
+          cpanm --quiet --notest Alien::Tidyp
+          cpanm --quiet --notest Net::SSLeay
+          cpanm --quiet --notest https://github.com/chapmanb/vcftools-cpan/archive/v0.953.tar.gz
+          
+          ./scripts/install_wsi_dependencies.sh "$NPG_LIB" \
+             perl-dnap-utilities \
+             perl-irods-wrap \
+             ml_warehouse \
+             npg_ml_warehouse \
```
mgcam, line 92: The current code is unlikely to depend on npg_ml_warehouse. st::api::lims::ml_warehouse, previously in that package, was moved to npg_tracking.

mgcam (changes requested):
Can we try dropping dependency on npg_ml_warehouse?

Otherwise looks good.

.github/workflows/run-tests.yml
kjsanger: Removed, still passes.

PR #803, opened 2023-10-05
Added a test to expose a problem with ref cache

mgcam:
Depends https://github.com/wtsi-npg/npg_tracking/pull/761


ces (approved):
Tests to show existence of ref cache bug so expected to fail until PR to resolve issue is made.

PR #805, opened 2023-10-18
Removed provisions for inline indexes.

mgcam:
Removed provisions for inline indexes in creating p4stage1 jobs. These indexes have not been used for the last 6 years.


dozy:
Looks fine. No corresponding changes to p4 templates required.

PR #815, opened 2024-01-05
Use one aggregation method for merges.

mgcam:
Use st::api::lims->aggregate_libraries() method for both 'merge_lanes' and 'merge_by_library' pipeline options.

This is a breaking change as far as archival and deletion of NovaSeq Standard flowcell is concerned, some forward planning is required.


dkj:
Key (only?) changes in output for a NovaSeq Standard (lane merging) will be: tag0 and tag 888 will not be merged over the lanes - is that right? This should definitely go in `Changes`...

I like the net reduction in code (taking npg_tracking change into account). 

The NovaSeq Standard workflow, where there is only one input port, is different to the more general merging across lanes where the (claimed) same library has been sequenced. But I don't think this is a reason to maintain separate code - there's no material positive.

PR #838, opened 2024-04-10
Extended the dir_path method of the product.

mgcam:
Wrapped around the inherited method in npg_pipeline::product so that it works for partial merges. Removed a flag that previously has to be passed to this method.

This fixes a bug in npg_pipeline::validation::entity - incorrect prediction of paths for partial merges.


dkj:
Nice discrete refactor!

PR #842, opened 2024-05-07
Archive Illumina samplesheet if present

mgcam:


dkj:
Hmmm - why did we not `--exclude Data` before?

PR #849, opened 2024-05-31
Changed mark duplicate method retrieval.

mgcam:
Ensured mark duplicate method can be inferred for a product with multiple studies.

Depends on https://github.com/wtsi-npg/npg_tracking/pull/833


dkj (approved):
ok from eyeball only

PR #876, opened 2025-01-15
tile limit flags for p4 stage1

dozy:
add flags to allow optional specification of first tile and tile limit for stage1 i2b


mgcam:
I am not sure how important this is, but the interop autoqc check result will not reflect tile selection.

PR #877, opened 2025-01-31
Restricted pipeline daemons to deal with runs by a certain manufacturer.

mgcam:
By default to runs on Illumina instruments.


marcomoscasgr (approved):
Pipeline deamon now supports the manufacturer name when picking up runs, which defaults to Illumina.

PR #895, opened 2025-07-28
add archive_irods_locations_to_ml_warehouse to analysis pipeline via function_list_central.json

ces:


mgcam:
We create a JSON file with locations unconditionally https://github.com/wtsi-npg/npg_seq_pipeline/blob/1541240b3ed2ba69ac6c24d5bfe24efda2b6822f/lib/npg_pipeline/function/seq_to_irods_archiver.pm#L52 so it is safe to slot in this function without changing any code

PR #899, opened 2025-10-01
Use minimap2 to align phix

jmtcsngr:


lib/npg_pipeline/base/options.pm
```diff
@@ -170,13 +170,7 @@ has q{p4s1_phix_alignment_method} => (
...
   my $self = shift;
 
-  my $alignment_method = $self->platform_NovaSeq? q[minimap2]: q[bwa_aln];
-
-  if($alignment_method eq q[bwa_aln] and not $self->is_paired_read) {
-      $alignment_method = q[bwa_aln_se];
-  }
-
-  return $alignment_method;
+  return q[minimap2];
```
mgcam: No need to have a build method for this attribute. Just set default value.

mgcam:
The charade of methods for detecting the phix reference in `p4_stage1_analysis.pm` should be reviewed and simplified.

Also, if `p4s1_phix_alignment_method` is not needed as a pipeline option, consider moving to `p4_stage1_analysis.pm`

mgcam (changes requested):

lib/npg_pipeline/base/options.pm
jmtcsngr: Thanks! Removing build method in next commit.

jmtcsngr:
> The charade of methods for detecting the phix reference in `p4_stage1_analysis.pm` should be reviewed and simplified.
> 
> Also, if `p4s1_phix_alignment_method` is not needed as a pipeline option, consider moving to `p4_stage1_analysis.pm`

Thanks @mgcam, commit e338df2 moves things around to p4_stage1_analysis, checked with @dozy and seems it is reasonable and easy to revert if we change our minds. I'm not sure bigger refactoring will help us further until we are sure we are happy with the change.

mgcam (approved):

PR #906, opened 2025-11-11
add runtime date stamps to viv logs, et al

dozy:
add runtime date stamps to viv logs, run json files and export_params files


dozy:
See https://jira.sanger.ac.uk/browse/NPG-3649

